PR #7, opened 2016-08-18
007-nsatz-inequalities: more efficient algorithm

andres-erbsen:
- The idea for the optimization is from https://github.com/coq/coq/pull/242#issuecomment-231128236 by @thery
- https://coq.inria.fr/bugs/show_bug.cgi?id=4880 is marked resolved, so it is probably no longer relevant to the CEP
- I am not sure whether the section about decidable equality is still relevant -- I think it is not, but I am not confident, so please check before this is merged.

@JasonGross: the performance of this algorithm could be tested out on https://github.com/JasonGross/fiat-crypto/blob/wip-weierstrass/src/WeierstrassCurve/WeierstrassCurveTheorems.v without implementing the field-of-quotients conversion because file already assumes that the integral domain in question is also a field.

### DIFF
--- a/text/007-nsatz-inequalities.md
+++ b/text/007-nsatz-inequalities.md
@@ -19,7 +19,24 @@ allow it to handle inequalities, and, if possible, be complete on that domain.
 
 Suppose we have a solver, `nsatz`, for polynomial equalities over an integral
 domain `F`.  We can reduce the following problems to already solved problems,
-thus enhansing `nsatz`:
+thus enhansing `nsatz` in two ways:
+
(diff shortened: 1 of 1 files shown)


JasonGross:
The removal of the reference to the resolved bug and the removal of decidable equality are good; thanks!

I'm not sure about the replacement involving the field of quotients (indeed, it would be best to do reflexively if at all).  Perhaps note both options, and leave as an open question which is easier / more efficient?

I'll aim to test this in the near future, but I don't think merging should block on that

PR #10, opened 2016-09-22
[009] Unified Pretty Printing RFC.

ejgallego:
We propose to enrich the current `Pp.std_cmdpps` datatype, consolidating
both `Ppextend` (thus adding notation support) and `Richpp` (thus adding
style/semantics support) into it; and making it serializable so it can
be interpreted client side.

### DIFF
--- a/text/009-unified-pretty-printing.md
+++ b/text/009-unified-pretty-printing.md
@@ -1,0 +1,167 @@
+- Title: Towards a new document type for Coq.
+
+- Driver: Emilio J. Gallego Arias (@ejgallego),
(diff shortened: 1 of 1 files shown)


ejgallego:
Thanks!

ejgallego:
I've added (@yurug) to the document. It would be great if he could comment more on the questions asked in coq/coq#185 . In particular, we knew of no user of the extended tagging infrastructure (*), so it would greatly help if he could describe how he's using it.

(*) SerAPI has now support for `PpAnn` as outlined in the document, but we were not happy with the results, it was however very instructive.

PR #14, opened 2017-01-24
Update 000-coq-enhancement-proposal.md

gares:


gares:
I'd like to implement the suggestion of @silene at the last CoqWG.
I tried to make the workflow more explicit.
Still I've no idea how to "hot swap" the current repo the the new model where "merging a CEP" means accepting it.

@silene do you have ideas?

silene:
I guess all the existing ceps could be moved to some `legacy-not-yet-accepted` directory so that they don't get lost, except for the `000` cep, which would then be the only accepted cep, by virtue of merging your pull request (assuming nobody notices an issue that prevents its merging). Then people would submit pull requests for either brand new ceps or for moving old ones from `legacy` to `text`, thus opening discussions about them.

silene:
By the way, to follow some existing practices, I would suggest that each cep should contain the following sections (possibly empty):

- summary
- motivation
- detailed design
- drawbacks
- alternatives
- unresolved questions
- updates since acceptation (for accepted ones)

text/000-coq-enhancement-proposal.md
```diff
@@ -31,16 +31,17 @@ a CEP may become irrelevant, obsoleted by the circumstances.
...
+1. summarize the outcome in the CEP, and keep the document updated
+1. defend it at the working group
+1. eventually write code or find someone that does it
 
 ## FAQ
-* Why not a pull request? A CEP is about the design, and has to happen before code is written
-* Why not a bug tracker entry? Why not a mailing list thread? A CEP can be related to a bug number and discussion can happen on a mailing list.  A CEP summarizes the discussion into a more accessible document
+* *Why not a pull request (to https://github.com/coq/coq with the code)?* A CEP is about the design, and has to happen before code is written. A proof of concept (a toy implementation) is perfectly fine and may help the discussion. Still, pull requests on https://github.com/coq/coq are meant for code to be integrated, not for ideas to be discussed.
+* *Why not a bug tracker entry? Why not a mailing list thread?* A CEP can be related to a bug number and discussion can happen on a mailing list.  A CEP summarizes the discussion into a more accessible document.  If you want to bring someone in the discussion a CEP is a single pointer to give to him.
```
Zimmi48: Advice: Use `them` instead of `him` (gender-neutral).

Zimmi48:
With this new workflow, we should make clear that once a CEP is accepted, editing it requires making a PR except to fix typos. This won't be very cumbersome because the main edition work will happen before the CEP is accepted / the is PR merged anyways.

JasonGross:
Would this mean that ceps should no longer be numbered?  Or that numbers are order of submission and not order of acceptance?

Zimmi48:
Probably they shouldn't be numbered. Or if we want to be able to mention CEP #number then we could use the #number of the PR which introduced it.

gares:
reopening it later

PR #16, opened 2017-03-03
Proof modes

gares:


text/ProofModes.md
```diff
@@ -0,0 +1,118 @@
...
+    foo.tac arg1.
+    bar.tac arg1.
+    ```
+    
+1. Linking (not even requiring) a plugin loads into the *lexer* the tokens 
+   specific to the grammar implemented by the plugin
+   
+1. Linking (not even requiring) a plugin loads into the *parser* grammar the rules
+   specific to the plugin 
+
```
ejgallego: Well, the "official" API here is to use `freeze/unfreeze` if you are not happy about the changes, right? It is indeed very messy.

text/ProofModes.md
```diff
@@ -0,0 +1,118 @@
...
+   
+   - `decl_mode`, for example, cannot use the `* EXTEND` mechanism because it needs to
+     plug in its own non-terminal (to hide ltac).
+
+1. So called "proof mode" is implemented by making the main "tactic" non-terminal
+   mutable and having specific commands assign such non-terminal to the desired one.
+   The current proof mode is not part of the `Summary` and is both stored in `Proof_global`
+   and `Stm` that do their best to ensure parsing happens in a state where such non-terminal
+   holds the expected value.
+
```
ejgallego: I had a look to this and apart from the difficulty navigating the code mess, I don't see a reason proof mode couldn't be kept in the summary. I am not sure _we_ want to do this, but it should work and relieve the STM from doing it. Another issue is that it seems to me that the proof mode handling in `proof_*` is indeed buggy, and what the STM is doing is effectively workarouding the bugs in that lower layer.

Ideally, IMHO we would just nuke most of `proof_*`

text/ProofModes.md
maximedenes: Is that API enough to implement support for the scenarios below?

text/ProofModes.md
```diff
@@ -0,0 +1,118 @@
...
+## Proposal
+
+Each proof mode installs all its grammar in a dedicated non-terminal `mode-root`.
+Such non-terminal is installed as the main entry (today the one for `tactic-expr`)
+in one of the following ways:
+- `Set Default Proof Mode "mode1".` (TODO: disable inside proof?)
+- `begin mode1. .... end.` within a proof.
+- `mode1:: ... .` within a proof, changes mode for 1 statement only
+- `Proof .... mode mode1 ... using ... .` (Useful?)
+
```
ejgallego: Looks good, [other than I still don't know how to make parsing clean wrt to these modes] What mode would be activated for proofs not starting with `Proof`?

text/ProofModes.md
ejgallego: Not without galaxy-sized hacks.

ejgallego: I just wanted to point out that as far as I understand the thing, it is not like it was not supported, but that the current official way is that you are supposed to do galatic hacks everytime you care about state in Coq.

text/ProofModes.md
gares: The only trade off between Summary and STM is the following one.

In Summary: to set the right parsing state, one has to install a full system state (no work on the STM side, since we do/should-do this already).

In STM: more work, but one could set the right parsing state without even having computed the corresponding system state.  One could only process "notation commands", i.e. be faster.

ejgallego: The more I think the more convinced I am that indeed this doesn't belong either in the summary or in the stm, but in a `parsing_state` type, which is handled by normal closures. Will try to summarize for the WG

text/ProofModes.md
Zimmi48: @ejgallego There has been a long time wish from some to make the use of `Proof` compulsory. Would there be any drawbacks doing this?

ejgallego: Good point, I am not sure how much it could be gained.

Zimmi48:
In the title of the CEP and in Maxime's talk, this proposal is advertised as "proof modes".

The general idea of having a built-in mechanism to implement various "proof modes" seems to me like a very nice idea. It could be used to make a better behaved declarative mode (if one wanted to do that job); maybe it could be used to avoid the current `MProof.` of MetaCoq; we can even think that it could be used for Iris.

But it isn't (as I understand it) what the CEP really discusses. The CEP is just about not loading lexer / parsing rules too early. The problem it tries to solve (in my limited understanding of this part of Coq) is pollution of the proof environment with unwanted tactic notations (when the user did not do `Require Import`).

Moreover, I fail to see SSReflect (the main application that you have in mind if I understood correctly) as a distinct proof mode. To reuse the terms of Pierre-Marie, SSReflect looks more like a tactic library (i.e. a set of tactics), which can be used in the classic proof mode, in combination with Coq built-in set of tactics.

Thus, instead of talking about proof modes, I would talk about a better behaved module system.

To continue talking about the application to SSReflect,  one could imagine putting Coq basic tactics into various modules which you need to import to use the tactics. Then you could `Require Import SSReflect.` and `Require Import BasicTacticsAlsoUsefulToSSReflectUsers.` Similarly, one could consider being able to do `Require Import 86Tactics.`

Instead of having to specify in which proof mode you are at the beginning of each proof, you could also consider having a `Local Import` feature. This would have the advantage over proof modes to be more generic.

You could also imagine putting all standard tactics into just one file but adding a feature to selectively import a few of them `Require Import Tactics [reflexivity, now, easy, discriminate]`.

Anyway, this is just food for thought. I still believe that the problems that you describe in this CEP are worth fixing.

ejgallego:
See also https://github.com/coq/coq/pull/459 the diff may help to understand more some basic first steps.

maximedenes:
> In the title of the CEP and in Maxime's talk, this proposal is advertised as "proof modes".
> 
> The general idea of having a built-in mechanism to implement various "proof modes" seems to me like a very nice idea. It could be used to make a better behaved declarative mode (if one wanted to do that job); maybe it could be used to avoid the current MProof. of MetaCoq; we can even think that it could be used for Iris.
> 
> EDIT: I might not be using the term "proof mode" for what it is supposed to mean, thus the confusion. From the talk, I understood "proof mode" in an intuitive way, without knowledge of this part of the implementation.

Yes, "proof mode" is an already existing terminology. It is used by the declarative mode, and this CEP is about refining (and among other things, implementing correctly) this mechanism.

> 
> But it isn't (as I understand it) what the CEP really discusses. The CEP is just about not loading lexer / parsing rules too early. The problem it tries to solve (in my limited understanding of this part of Coq) is pollution of the proof environment with unwanted tactic notations (when the user did not do Require Import).

The CEP is not only about not loading rules too early. It is also about a fine grained mechanism to switch between modes, and the possibility to use libraries (i.e. to Require .vo files) independently of the language (mode) they were written in.

> 
> Moreover, I fail to see SSReflect (the main application that you have in mind if I understood correctly) as a distinct proof mode. To reuse the terms of Pierre-Marie, SSReflect looks more like a tactic library (i.e. a set of tactics), which can be used in the classic proof mode, in combination with Coq built-in set of tactics.

I don't understand the distinction you are trying to make. First, I'm very surprised that you don't see SSR as a language. Is it a feeling that you got by using it? In my view, it is a proof language whose atomic blocks are tactics. Can you make more explicit what you would call a proof mode? And a proof language?

For example, the declarative mode does nothing more than importing its own tactics and syntax. Formally, Iris is the same, except that it acts on embedded contexts and domain specific logics (but that's part of the tactics implementation).

Second, Ssreflect is typically *not* an application of this framework, because it was *designed* to have no syntactic conflicts with the 8.6 tactics (apart from a few places like rewrite where one can still disambiguate without resorting to proof modes). What this CEP brings is more freedom for others to implement other proof languages when we merge SSR.

Zimmi48:
> Yes, "proof mode" is an already existing terminology. It is used by the declarative mode, and this CEP is about refining (and among other things, implementing correctly) this mechanism.

I was confused indeed, so most of my earlier comments stemmed from this confusion. 

> I don't understand the distinction you are trying to make. First, I'm very surprised that you don't see SSR as a language. Is it a feeling that you got by using it? In my view, it is a proof language whose atomic blocks are tactics. Can you make more explicit what you would call a proof mode? And a proof language?

Don't get me wrong. I never said SSReflect was not a proof language. I already said that it is a well-thought, consistent set of tactics and as such it can be called a proof language.
What I am saying is that it is much less different from the classic proof mode than the declarative mode is. See on this example:

```
Goal forall n, n + 0 = n.
proof.
  let n : nat.
  per induction on n.
  suppose it is 0.
    hence thesis.
  suppose it is (S n') and (n' + 0 = n').
    have (S n' + 0 = S (n' + 0)).
                  ~= (S n').
    hence thesis.
  end induction.
end proof.
Qed.

Goal forall n, n + 0 = n.
Proof.
  induction n.
  - reflexivity.
  - now simpl; rewrite IHn.
Qed.

Require Import mathcomp.ssreflect.ssreflect.

Set Bullet Behavior "Strict Subproofs".

Goal forall n, n + 0 = n.
Proof.
  elim.
  - reflexivity.
  - by move => n /= ->.
Qed.
```

*You will excuse me from putting back the default behavior of bullets. I can't stand not having them. By the way, I hope that you plan to take advantage of the SSR 2.0 release to make the standard bullets usable with SSReflect.*

> For example, the declarative mode does nothing more than importing its own tactics and syntax. 

Does it? OK, but it feels like it does more. Indeed, as can be seen by executing the previous example step by step, the "tactic" `per induction on n.` does not affect the proof term, nor the goal to prove, but it does some focusing and changes some hidden state so that it is expecting tactics of the form `suppose it is ...`

> Second, Ssreflect is typically *not* an application of this framework, because it was *designed* to have no syntactic conflicts with the 8.6 tactics (apart from a few places like rewrite where one can still disambiguate without resorting to proof modes). What this CEP brings is more freedom for others to implement other proof languages when we merge SSR.

Thanks for the clarification! This was indeed something that I had not understood so far.

Anyway, do not take any of my comments as criticism. I feel that you are proposing something useful here, although I don't have enough knowledge of the system to judge.

ejgallego:
> Goal forall n, n + 0 = n.
> Proof.
>   elim.
>   - reflexivity.
>   - by move => n /= ->.
> Qed.

By the way the correct proof here is
```
Goal forall n, n + 0 = n.
Proof. by elim. Qed.
```
if you want to be more explicit you could do `by elim=> // n ->.`

> I hope that you plan to take advantage of the SSR 2.0 release to make the standard bullets usable with SSReflect.

This has been already discussed, indeed the "standard" bullets are non usable in many proofs, unless you want to indent like up to tab 80. Bullets were first done in SSR then copied in an incompatible way to Coq.

Anyways, I am the first to blame wrt leading PRs off the tangent, but all this discussion about tactic/proof languages starts to be IMVHO a bit tiring, specially given that often comments about the SSR tactic language come from people who has barely used it so they are not particularly useful.

Tactic discussion is important, but maybe the right place to keep things organized would be a thread on coqdev. Cheers!

Zimmi48:
> By the way the correct proof here is

Come on, there is no such thing as "the correct proof". Anyways, `by elim=> // n ->.` is not making things more explicit, it is making things less explicit since `elim => //.` or even `move => //.` is actually enough to finish the proof.

> This has been already discussed, indeed the "standard" bullets are non usable in many proofs, unless you want to indent like up to tab 80. Bullets were first done in SSR then copied in an incompatible way to Coq.

You're kidding me right? This is not by taking such positions that you are going to convince people to use SSReflect. SSReflect's bullets are decorative (and counter-intuitive when you know Coq's bullets). I don't care about decorating my proofs.

> Anyways, I am the first to blame wrt leading PRs off the tangent, but all this discussion about tactic/proof languages starts to be IMVHO a bit tiring, specially given that often comments about the SSR tactic language come from people who has barely used it so they are not particularly useful.

Yes you are, so I don't feel bad replying and feeding the troll. Maxime and Enrico's claim is that SSReflect 2.0 is a general purpose language which could be used by anyone doing proofs with Coq and that by keeping it out of the main codebase, we are driving users away (e.g. CompCert apparently). They also claim that this is a solution to compatibility issues and thus that it could solve the problems that some users have. Given such strong opinions (which I would call expansionist), I am considering opening myself to SSReflect (not the style, but the language) but this kind of reply that you are giving me is clearly pushing me out.

Since we are at the level of troll, what about this thing of not putting spaces between a tactic and `=>`. You were the one (I recall) telling that more spaces make the reading easier (in OCaml). Why doesn't it apply to proofs as well? Or maybe readability is not a goal...

> Tactic discussion is important, but maybe the right place to keep things organized would be a thread on coqdev. Cheers!

Yes, you're right. I would have felt bad if I hadn't answer to your comments but really, we are just polluting the discussion on this CEP and I feel bad for Maxime and Enrico. This is not the place to discuss SSReflect. Let's drop the subject, alright? Cheers!

ejgallego:
> Come on, there is no such thing as "the correct proof".

I am not the best person to comment but indeed in math-comp there is a proof style, and I am pretty sure that `by elim` would be the correct style here. Indeed, there is a correct style because the authors of the library chose so.

> You're kidding me right? This is not by taking such positions that you are going to convince people to use SSReflect. SSReflect's bullets are decorative (and counter-intuitive when you know Coq's bullets). I don't care about decorating my proofs.

I am not kidding you. First, I don't care about what people use and I am not trying to convince anyone, as it is each person work to realize what is best for them. I am just pointing out a very serious problem with the "standard" bullets, you can read the summary here: https://github.com/math-comp/math-comp/issues/102 Secondly, I am complaining about your introduction of a totally unrelated statement here "bullets" that pollutes the discussion, and it seems it is not very informed from the practical or historical perspective. Thirdly, I must apologize because I do that all the time in GH.

I am doubly sorry if you considered my reply aggressive and I apologize if I am pushing you out. My intention was just to correct several statements that were verifiably not correct.

There are quite a few other statements in your reply that are easy to rebuke, but I will refrain from doing so as it don't think it is productive, but on a small ending note, the style between a tactic and `=>` is specified in the ssr manual, page 60 https://hal.inria.fr/inria-00258384v16/document

text/ProofModes.md
ejgallego: A slight technical point here is what to do when a plugin is "unloaded", that is to say, imagine the plugin successfully registers its tables, now when we go to the part of the document before the plugin was loaded, should these tables be dropped too?

With my current design they will, as the tables themselves are part of the parsing state, thus, we may need to be sure that these tables are registered even when the "already loaded" optimization is done.

The other alternative is to make these tables global, but again, it would be good if we could avoid that.

ejgallego:
Ok, so I've updated coq/coq#459 to a point where it is almost ready to handle the proper extends. We can continue the discussion about code there.

text/ProofModes.md
```diff
@@ -0,0 +1,118 @@
...
+In this case we don't get why the second rule does not mask the first one.
+
+```
+Notation "! 'bar' x " := 1 (at level 2).
+Notation "! bar x" := x (at level 2).
+
+Goal ! bar 2 = 3.
+```
+
```
herbelin: `Print Grammar constr` shows:
```
...
| "2" RIGHTA
  [ "!"; IDENT "bar"; NEXT
  | "!"; constr:operconstr LEVEL "200"; NEXT ]
...
```
This indicates that both rules are active (and overlapping only on expressions of the form `! bar any-term`). It seems that the convention of camlp5 is to give priority to the rule which parses a terminal, so here trying to parse the terminal `IDENT "bar"` before trying to parse the non terminal `operconstr`.

text/ProofModes.md
```diff
@@ -0,0 +1,118 @@
...
+Also (this time it prints 'bar' is a keyword).
+
+```
+Notation "'bar' ! x " := 1 (at level 2).
+Notation "bar ! x" := x (at level 2).
+
+Goal bar ! 2 = 3.
+```
```
herbelin: An ident has to be declared as a keyword whenever it could be absorbed by the non-terminal rule which precedes it.

In `! 'bar' x`, there is no need to make `bar` a keyword since it is preceded by a terminal. In `'bar' ! x` however, the expression can occur as part of the rule parsing application, typically `operconstr operconstr`, so it can be preceded by the non-terminal `operconstr` which itself can parse a trailing `IDENT` since any `IDENT` is also an `operconstr`. So, to ensure, say that `f bar ! g`, which is parsed from the left without deep backtracking (my vague memory is that camlp5 is something like [LL(1)](https://en.wikipedia.org/wiki/LL_parser) + some extra arbitrary long look-ahead for terminal), is not seen as the application of `f` to the variable `bar` then to something else yet to be parsed, `"bar"` has to be declared as a keyword.

Is that clear enough?

herbelin:
I have a question: is this CEP supposed to cover the managing of different versions of a same tactic, like the version of `apply` which uses `w_unify` and the version of `apply` which uses `evarconv`?

ejgallego:
@herbelin maybe not completely but it would certainly help.

herbelin:
@ejgallego: do you have already some model in mind?

I don't know the exact plans of @mattam82, but typically, we will have 
- either an ML function `apply` which at some time queries a flag to decide whether it uses `Meta` and `w_unify` or `Evar` and `evarconv`
- or to distinct ML functions `apply1` and `apply2`, one using `w_unify` and the other using `evarconv`

Now, users will probably have some old files which I guess they would like to compile with the legacy `apply` and new files for which they would like to take benefit of the stronger `apply`. How to manage this? What are the different effective models that we have at our disposal to implement this?

Zimmi48:
@herbelin Enrico's suggestion as explained here https://github.com/coq/coq/pull/539#issuecomment-291877985 was to duplicate the code, move the new code to `evarconv`, give access to both implementations thanks to two modes and remove completely the old code in an upcoming version. It looks to me like a good suggestion because it will avoid unnecessarily complicating the new code, although it will lead to temporary code duplication. Of course, to avoid paying the price of the duplication, after the release bugs should only be fixed in the new version.

herbelin:
@Zimmi48: Factorization of code vs duplicated code is a real question. Both have advantages and drawbacks.

@ejgallego: The possibility to have attributes on any node of a syntax tree (this is my understanding of  coq/coq#402) is powerful. For tactics, I suspect we would need attributes only on a very few nodes but why not using a general infrastructure for attributes after all.

mattam82:
@ejgallego @herbelin Indeed I'd also prefer the complete separation of the code between the two versions of e.g. apply. I don't see why the proof-modes wouldn't be enough here, at least with the global Set Default Proof Mode Coq-8.6 / Coq-8.7 and a finer grain tactic level (mode 8.6: apply foo), do you think that the latter is unlikely to happen? Do you suggest this finer grain mode/version switching could be implemented with attributes?

ejgallego:
@mattam82 proof modes work, however I am concerned about scalability, in the sense that in the current state, you need to have a different proof mode for each tactic, which could quickly lead to proliferation of proof modes.

mattam82:
Hmm, why couldn't you attach a set of tactics to a single proof mode?

ejgallego:
Yes you could. But we may really need "tactic modes" and "proof modes", in the sense that people may want to have different tactic language modes (LTAC / LTAC 2) and different selection of tactics, imagine:
```
(apply H)%v8.7 ; (apply M)%v8.6
```
and this kind of tactic selection is really specific to the tactic language syntax.

Zimmi48:
Honestly, I think it should be bearable if the proof mode can only be set at the sentence level.
Especially for the 8.7 vs 8.6 mode as this is only for compatibility during the transition.

ejgallego:
@Zimmi48 yup, it would be good anyways to factor the ltac grammars a bit.

herbelin:
@mattam82: What are the steps to arrive to this model? Does the plan start with something like separating the current directory `plugins/ltac` into `plugins/ltac/v8.6`, `plugins/ltac/v8.7`, `plugins/tactics/v8.6`, `plugins/tactics/v8.7`? Duplicating tactics.ml towards `plugins/tactics/v8.6` and `plugins/tactics/v8.7`? Moving g_tactics.ml4 to both `plugins/tactics/v8.6` and `plugins/tactics/v8.7`? Etc.

herbelin:
I'm a bit lost in the discussion. What is the status of proof modes? Is it already usable today for one or another purpose?

PR #18, opened 2017-03-16
A versioned, independent, development model for the standard library

Zimmi48:
Summary
=======

Versioning the standard library independently from Coq itself is important for better compatibility, especially in a development model where new versions of Coq are quite frequent.

For practical reasons, this is better done by moving the `theories/` files out of the main repository and into a specific coq/lib repository.

This move will also be a perfect opportunity to make sure that users are encouraged to contribute to the improvement of the library, and possibly to find additional maintainers coming from the user community.

--------

I wanted to start this CEP after a discussion about compatibility with @herbelin and @ejgallego on https://github.com/coq/coq/pull/140#issuecomment-286724361.

It draws inspiration from demand of users such as Michael Soegtrop. I also had some very preliminary discussions with e.g. @amahboubi.

I had a first review of this CEP offline with Hugo, Matthieu and Daniel, but this is still in a very preliminary state. I would appreciate your comments in order to make it better.

### DIFF
--- a/text/000-versioned-stdlib.md
+++ b/text/000-versioned-stdlib.md
@@ -1,0 +1,162 @@
+- Title: A versioned, independent, development model for the standard library
+
+- Drivers: @Zimmi48 (co-drivers could be @ejgallego @herbelin @mattam82 @amahboubi ...)
+
+- Status: very preliminary draft, RFC
+
+----
+
+# Summary
+
+Versioning the standard library independently from Coq itself
+is important for better compatibility, especially in a development
+model where new versions of Coq are quite frequent.
+
+For practical reasons, this is better done by moving the `theories/`
+files out of the main repository and into a specific coq/lib repository.
+
+This move will also be a perfect opportunity to make sure that users
+are encouraged to contribute to the improvement of the library,
+and possibly to find additional maintainers coming from the user community.
+
+# Motivation
+
+The main motivation is to address the compatibility concern expressed
+by lots of users, at the level of the library. Moving to an independent
+versioning scheme will allow users to upgrade Coq without changing their
+version of the library; and to upgrade their dependency on the library
+without changing their version of Coq.
+
+# Detailed design
+
+A new repository coq/lib is created which contains the current state of
+the `theories/` files (except those which are absolutely necessary for
+building Coq).
+
+The development of the library continues in this repository and respects
+semantic versioning.
+
+## Semantic versioning means:
+
+- major versions can introduce breaking changes: they can add and remove
+things, change definitions, change names of theorems;
+
+- minor versions are about non-breaking improvements: they can only add
+new definitions and theorems; adding a new theorem to a hint database is
+considered a minor change: if users' scripts are relying on the failure
+of an automatic tactic (except if this tactic implements a specific
+decision procedure), it is completely fragile and thus it is their problem;
+
+- patch level versions are about fixing bugs: bugs are deviations between
+specification and implementation in definitions and decision procedures;
+normally if enough theorems are proved about a definition, it will
+probably not be buggy and thus patch level versions should be quite rare.
+
+## Development practices
+
+There is one branch per released major version (the stable branches, which
+are named v1, v2, v3...) plus the master branch.
+
+Breaking changes are always done on the master branch only.
+
+Non-breaking changes are always done on the latest stable branch, then
+merged into the master branch.
+
+All the supported versions of the library (the development version, the
+stable version and possibly a few legacy versions, i.e. older stable
+versions) are checked to compile with all the supported versions of Coq
+(thanks to automatic testing jobs).
+
+## Distribution
+
+The latest stable version and possibly some legacy versions are directly
+included in the Coq packages which are distributed for Windows, MacOS and
+Linux distributions.
+
+All the other versions, including the development version, are accessible
+through OPAM packages, or by manually downloading and compiling the
+corresponding archives.
+
+## Call for contributors
+
+Once the split is done, it would be a good thing for the library to have
+its own webpage and motivated maintainers who spread the word that it is
+open to contributions from the Coq user community.
+
+It can be discussed later but maybe some library developers would like to
+integrate part of their work directly in the standard library.
+
+The models here could be the QED project and the archive of formal proofs
+of Isabelle (and contribution models of other provers, e.g. Mizar, ACL2...).
+
+# Drawbacks
+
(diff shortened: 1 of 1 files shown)


psteckler:
So this would mean that if I distribute a Coq development, I need to specify both a Coq version and a Coq-lib version that are compatible (or perhaps multiple versions of those)?

Zimmi48:
@psteckler That's an interesting question with two levels. First, what to specify as dependencies. Second, how to tell Coq which version of the library to use.

On OPAM, we could consider having a package for Coq itself (without plugins and without any library) and packages for plugins and the library. Then, when distributing a development via OPAM you could use the existing dependency management system to specify bounds on the required versions of Coq and the library.

Unfortunately, OPAM does not handle (as far I know) having several versions of the same package installed in parallel (unless you use different switches or roots) which could make it a bit cumbersome to switch from one version of the library to another (although it is not quite clear that this would be something that users would do often).

You also need a way to be able to tell Coq which version of the library to use since the CEP says that we would be shipping several versions of the library in a single non-OPAM Coq package. That could be managed at the level of `coqtop` options, or it could be done with a command inside the document itself. I wonder what people think about that.

psteckler:
Is there a place in `_CoqProject` files to specify such dependencies? Or maybe that's what you meant by `coqtop` options.

Zimmi48:
Yes, precisely!

silene:
Among the unresolved questions, you mention the issue of the `plugins` directory, but you are underestimating the issue. For instance, `theories/Reals/RIneq.v` depends on `plugins/setoid_ring/Ring_tac.v`, which depends on `theories/Reals/Rdefinitions.v`. So the dependency between plugins and the standard library is not just one way, it is both ways. Moreover, this is a checksum-based dependency!

As a consequence, I believe this CEP is just a sweet dream until one actually exhibits two Opam packages for the Coq binaries and two Opam packages for the standard library, and shows that both Coq packages are compatible with both Stdlib packages. If they are tightly tied in practice, making the split is pointless.

Finally, I think that splitting by components (as proposed by @letouzey in coq/coq#498) is much saner, both in the short run and in the long run. Yet, as with this CEP, one of the (unmentioned) unresolved question is the transition period. How does one make an external development compatible with both a shipped standard library and a separate standard library?

Zimmi48:
@silene
> So the dependency between plugins and the standard library is not just one way, it is both ways.

Yeah I know, and the more I thought about this, the more I thought that some plugins should move along with the library. Although I'm pretty sure that the other approach is still doable, with a few dynamic checks in the plugin code.

> Finally, I think that splitting by components (as proposed by @letouzey in coq/coq#498) is much saner, both in the short run and in the long run.

This CEP isn't incompatible with splitting the library into several parts. I would say it is even orthogonal to @letouzey's proposal since one consequence of this PR is that we stop shipping some parts of the library in the main Coq packages, while this CEP proposes that all the parts that are moved to this new repository continue to be shipped in the main Coq packages.

Besides, the splitting approach won't help much for improvements such as the one proposed in coq/coq#140 as this is touching one of the most basic brick of the stdlib.

> Yet, as with this CEP, one of the (unmentioned) unresolved question is the transition period.

If I understand correctly, OPAM dependency declarations are expressive enough to be able to say: with Coq 8.6 or earlier, I don't need any other package, but with Coq 8.7 or later I need the "stdlib" or the "bignum" package:

    ("coq" & "stdlib") | "coq" {>= "8.4" & <= "8.6"}

silene:
> If I understand correctly, OPAM dependency declarations are expressive enough to be able to say

My comment was not related to Opam. Before the split, a `.v` file contains

    From Coq Require Import Lists.

After the split, it contains

    From Whatever Require Import Lists.

How do you write your `.v` file so that it is compatible with both versions?

Zimmi48:
I would like to keep `From Coq Require Import Lists.` This is after all, still the standard library of Coq. As for which version of the library to use, I was proposing an option for coqtop (see the earlier discussion with Paul).

That being said, the problem you are highlighting is much more universal than just this CEP. Indeed, it applies as well to any splitting proposal. But it applies also to any integration proposal: i.e. now you do `From mathcomp Require Import ssreflect.` and tomorrow you would have to do `From Coq Require Import ssreflect.` I don't think that this kind of problem should prevent us to move forward though. When it is really important to have the same version of a package building with different versions of Coq, tricks can be used. One trick would be to put all the necessary imports in a special file, which is dependent on the version and is symlinked at make time.

ejgallego:
> `From mathcomp Require Import ssreflect`. and tomorrow you would have to do `From Coq Require Import ssreflect.`

Why would that happen?

Zimmi48:
@ejgallego Well, because @gares and @maximedenes proposed at the last WG to have SSR 2.0 be integrated into Coq (somewhere around https://youtu.be/_nki0UkUqoo?t=2h46m20s). Since they said it was just the tactic language, and not the math-comp library, I suppose we can't keep the same base name. Though, if I misunderstood, let's discuss this elsewhere (e.g. Gitter) and not pollute this thread.

ejgallego:
Well, what "integrated with Coq" means is becoming a bit fuzzy as this PR highlights. I think we may want to state more accurately "officially _distributed_ with Coq"

Anyways, even if SSR2 is part of the official Coq distro any day, IMHO it makes sense to still use namespaces. [In particular, mathcomp has very generic module names such as `seq` etc...]

silene:
> and tomorrow you would have to do `From Coq Require Import ssreflect.`

No you won't. Mathcomp would just have to provide a dumb `ssreflect.v` file which exports the plugin. That would be transparent for the user.

Zimmi48:
@silene Yeah, that's an idea. But anyways, my point was just to say "There will be necessary adjustments but this won't really be a big problem, you can always find a trick around these."

silene:
> this won't really be a big problem, you can always find a trick around these

As I said, you are underestimating the issues involved by such a split. No one has discovered the trick you mention yet.

ejgallego:
Indeed I like this proposal but as Guillaume says it is quite tricky to implement. I tried to do something similar in the past but I failed. However, I did succeed in splitting coqide and coqchk to their own separate repositories in a kinda clean way (and by the way the coqide split brought good advantages in the form of simplification of build logic)

I am wondering if maybe we could try to complete those splits first, I think that some of the changes needed would indeed be of help to the implementation of this proposal (build/test system)

Zimmi48:
Closing as this is somehow superseded by the Stdlib 2.0 project which @maximedenes is going to lead.

PR #20, opened 2017-06-13
RFC: Ltac2

gares:
Submitted in place of @ppedrot (feel free to take over this PR)

### DIFF
--- a/text/typed-ltac.md
+++ b/text/typed-ltac.md
@@ -1,0 +1,555 @@
+- Title: Ltac 2.0
+
+- Proposer: Pierre-Marie
+
+----
+
+# Summary
+
+The Ltac tactic language is probably one of the ingredients of the success of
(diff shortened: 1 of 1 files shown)


herbelin:
It seemed to me that there is already very cool and useful significant creative work done about Ltac2. Naive question: would it make sense to already integrate some parts of it, including datatypes and typing, with, maybe some hooks to reuse missing primitives from Ltac?

gares:
Maybe the title was misleading (fixed now) but this PR is really about the Ltac2 CEP.

Zimmi48:
Should this get merged? Closed?

ppedrot:
I think the standard course is to merge it, but it's kind of a duplicate with the Ltac2 documentation.

Zimmi48:
CEPs are snapshots of a design document at one point in time, whereas the documentation will evolve with the code.

PR #21, opened 2017-06-13
RFC: import-name-modifiers

gares:
@JasonGross  feel free to take over this PR


SkySkimmer:
High level semantics are important, but AFAICT we also need to redesign the libobject stuff to make partial import possible.
Such a redesign should consider other related feature requests, eg the local Import one (we don't have to grant all of them ofc).

PR #28, opened 2017-06-13
RFC: numeral-notation

gares:
@letouzey  feel free to take over this PR (related to https://github.com/coq/coq/pull/496 https://github.com/coq/coq/pull/719)

### DIFF
--- a/text/numeral-notation.md
+++ b/text/numeral-notation.md
@@ -1,0 +1,124 @@
+- Title: Numeral Notation
+
+- Driver: Daniel
+
+- Status: Draft
+
+-----
+
+# Reference
+
+  - [PR#156](https://github.com/coq/coq/pull/156)
+  - [PR#496](https://github.com/coq/coq/pull/496)
+
+# Summary
+
+Numeral Notation is a new proposed vernac command. It allows any Coq user to define his own numeral types to be parsed and printed as strings of decimal digits, possibly prefixed by a minus sign (such as "0", "-42", "825"). There is an implementation of that, but before integrating it into the Coq sources, there are several decisions to take. But let us see what it is.
+
+First, the type of "strings of decimal digits" has been named Z' (Z prime, there is an debate about this choice, see below). The vernacular command is the following:
+
+```
+Numeral Notation t f g : t_scope { (printing** <list of references>) } { (warning after <number>) }
+```
+
+* t is the type,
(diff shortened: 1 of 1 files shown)


text/numeral-notation.md
```diff
@@ -0,0 +1,124 @@
...
+
+ * Speed of parsing and printing for the other types (nat, Z, R) should be tested, but it depends whether Coq programmers use really big numbers of these types; if not, we could have this version which is probably slower that the previous one, but small compared to the other parts of Coq (e.g. tactics).
+
+ * Pierre Letouzey : I suggest to use a very simple base-10 encoding as Coq entry point of Daniel's new parsing/printing mechanism, and let Coq do the rest of the conversion between these base-10 numbers and nat/positive/N/Z/. A [proof-of-concept](https://github.com/letouzey/baseconv) is on github.
+
+  * Interest:
+
+   * Less uncertified critical code : during parsing at the OCaml level, we could keep numeral constants as strings (e.g. "123"), and simply convert these strings to Coq lists of base-10 digits, i.e. D1::D2::D3::nil in my first datatype proposal, see [Deci.v](https://github.com/letouzey/baseconv/blob/master/Deci.v), or (D1 (D2 (D3 Stop))) in the second datatype proposal in [DeciBis.v](https://github.com/letouzey/baseconv/blob/master/DeciBis.v).
```
SkySkimmer: Why not go further?
The above:
```
user string (ocaml value) =[ocaml]=> list digit (Coq value) =[coq]=> one of nat, Z, etc (Coq value)
```

More general:
```
user string (ocaml value) =[ocaml]=> list ascii (Coq value) =[coq]=> one of nat, Z, list digit, etc (Coq value)
```

JasonGross: How do you propose determining what is a numeral and what is not?  I believe the current code does not change the parsing of numerals, only the interpretation.  When I modified the code to strip off leading zeros for backwards compatibility, there was a noticeable performance impact when I did the stripping in too many places; this code is already somewhat performance sensitive.

If we're not allowing the parsing of strings of anything other than decimal digits, then allowing `list ascii ` seems like useless generalizing that will make the user do more work.

Zimmi48:
Is this time to merge this CEP? Close it?

gares:
I did not follow numeral notations. Did the PR that was merged follow this design?

JasonGross:
Roughly, yes.  The bit about Ltac and reference printing never got merged, we support conversion from BinInt.Z as well as Decimal.(u)int, but overall I believe the design is the same.

PR #30, opened 2017-08-31
My wishes for typeclasses

CohenCyril:


text/typeclass_control.md
```diff
@@ -0,0 +1,19 @@
...
+----
+
+# Wishes
+
+- A cut tag between the (class) arguments of instances, with the same behavior as the prolog cut (as opposed to the Hint Cut). Syntax TBD
+E.g. (using "cut" as a placeholder for the future syntax)
+```coq
+Instance A `{C x} "cut" `{C' x y} : C (f x y) := ...
+```
+If a solution for (C x) is found, then no backtrack is allowed if a search for (C' x y) fails.
```
silene: If I understand correctly what you wish, it seems like this could also be expressed as an additional mode for `Hint Mode`, since it already specifies which arguments are input-only.

CohenCyril: As of today `Hint Mode` describes the status of the indexes of a **Class**.
"Cut" is a marker in a prolog clauses i.e. **Instances**, for which the number of arguments is independent with the number of indexes of a class.
So no, `Hint Mode` is not appropriate.

CohenCyril: It's really a prolog cut as in https://en.wikibooks.org/wiki/Prolog/Cuts_and_Negation

ppedrot: FTR, you can encode a Prolog cut by directly writing a Hint Extern and using the `once` tactical over the set of typeclasses you want to cut on. It's very heavyweight though.

CohenCyril: I tried and failed, although I couldn't determine whether it was due to a mistake on my behalf or because the overhead is so heavy that it crashes... Debugging was so unreadable that I gave up.
Moreover, I am not sure what the interaction between unfolding and hints is.
If you have such an encoding, please share it with me so I can try yours :)

mattam82: @aspiwack Could we implement a cut on top of the engine actually, as a primitive ? Otherwise we need a kind of global setup for each instance, doing a kind of implicit "once" on the whole previous hypotheses

CohenCyril: However, my encoding had to be a bit more subtle than what you said because I do not want to cut arguments based on the typeclasses, but rather on the same typeclass but different positions of the arguments in an instance, so my encoding relies on an dummy typeclass with a Hint Extern which calls once typeclasses eauto...

ppedrot: Here you go:
```coq
Class A := { a : nat }.
Class B := { b : nat }.

Arguments a _ : clear implicits.
Arguments b _ : clear implicits.

Class C := { ca : nat; cb : nat }.

Instance a₀ : A := { a := 0 }.
Instance a₁ : A := { a := 1 }.

Instance b₀ : B := { b := 0 }.
Instance b₁ : B := { b := 1 }.

Ltac instance :=
  (** Block instance resolution on C in the refine below *)
  change (id C);
  let x := fresh in
  unshelve refine (let x := _ in _); [exact A | | ];
  [once typeclasses eauto|];
  let av := eval unfold x in x in
  clear x;
  unshelve refine (let x := _ in _); [exact B | | ];
  [typeclasses eauto|];
  let bv := eval unfold x in x in
  clear x;
  refine (Build_C av.(@a) bv.(@b));
  change C
  .

Ltac list_instance A :=
  try (
    let x := fresh in
    unshelve refine (let x := _ in _); [exact A | | ];
    [typeclasses eauto|];
    let ans := eval unfold x in x in
    clear x;
    idtac ans;
    fail
  ).

Hint Extern 2 C => instance : typeclass_instances.

Goal True.
list_instance A.
list_instance B.
list_instance C.
```

mattam82: Another solution would be to have a modifier for using once on each argument but that's not as powerful as cut.

CohenCyril: Ok, I tried again and it almost work (without `once` though), and here I am:
```coq
(**************)
(* Prolog cut *) 
(**************)

Fact cut_key : unit. Proof tt.
Definition locked_with : unit -> forall {T}, T -> T := fun 'tt => @id.
Class cut (class : Type) := cut_field : locked_with cut_key class.

Section cut.
Context (class : Type).
Local Notation cut := (cut class).
Lemma cutE : cut = class.
Proof. now unfold cut; case cut_key. Qed.
Definition put_cut : cut -> class.
Proof. now rewrite cutE. Qed.
Definition get_cut : class -> cut.
Proof. now rewrite cutE. Qed.
End cut.

Notation "'cut c" := (@cut c) (at level 200).

Require Import Classes.Init.
Hint Extern 0 (@cut _) =>
  (apply (@get_cut _ _)) : typeclass_instances.

(* Testing prolog cut *)

Class C (n : nat) (m : nat).
Hint Mode C + - : typeclass_instances.
Instance a : C 0 1 | 1.
Instance b : C 0 0 | 2.
Instance d x y `{'cut C 0 x} `{C x y} : C 3 3.

Goal C 3 3.
Set Typeclasses Debug. 
Fail typeclasses eauto.
Abort.
```
The big problem here is that apply breaks the flow of the proof search and the traces start back to 1., which means that debugging becomes unreadable because of the interleaving between the numerous proof search that are printed in a random order. Moreover any options of typeclasses eauto are lost.

But if I replace `apply (@get_cut _ _)` by `once (class_apply get_cut)`, it does not work. The test which is supposed to fail (because the first solution that works is "a" and thus "b" should never be considered) now succeeds :(

text/typeclass_control.md
```diff
@@ -0,0 +1,19 @@
...
+
+- Control typeclass opacity by class and not globally. The opacity/transparency constraint should be stored in a hint database in order for it to parametrize typeclasses eauto with ...
+
```
mattam82: One way to do this would be to have one database per class, with the option of putting instances in multiple databases, is that your next wish? We could also associate a transparency state to each class of course (to each head in the hint database).

CohenCyril:
> Could we implement a cut on top of the engine actually, as a primitive ?
Isn't `once` the primitive ?

text/typeclass_control.md
CohenCyril: If I do this, it works now :
```coq
Hint Extern 0 (@cut _) =>
  (eapply @get_cut); once typeclasses eauto : typeclass_instances.
```

ppedrot:
In Prolog syntax, `x :- a₀, ..., aₘ, !, b₀, ..., bₙ` becomes Ltac `once (a₀; ...; aₘ); b₀; ...; bₙ`.

CohenCyril:
In the end for a cut after the first argument it works with:
```coq
Class cut (class : Type) := Cut {cut_field : class}.
Notation "'cut c" := (@cut c) (at level 200).
Hint Extern 0 (@cut _) =>
  eapply @Cut; once typeclasses eauto : typeclass_instances.
```

But for a cut after n arguments, we need more encoding... (_affaire à suivre_)

ppedrot:
@CohenCyril No, that's just a generalization of the code I gave.

The code below encodes `X :- A, B, !, C`.
```
Class A := { a : nat }.
Class B := { b : nat }.
Class C := { c : nat }.

Arguments a _ : clear implicits.
Arguments b _ : clear implicits.
Arguments c _ : clear implicits.

Class X := { xa : nat; xb : nat; xc : nat }.

Instance a₀ : A := { a := 0 }.
Instance a₁ : A := { a := 1 }.

Instance b₀ : B := { b := 0 }.
Instance b₁ : B := { b := 1 }.

Instance c₀ : C := { c := 0 }.
Instance c₁ : C := { c := 1 }.

(** X :- A, B, !, C *)
Ltac instance :=
  (** Block instance resolution on C in the refine below *)
  change (id X);
  let x := fresh in
  let y := fresh in
  let z := fresh in
  once (
    unshelve refine (let x := _ in _); [exact A | | ];
    [typeclasses eauto|];
    unshelve refine (let y := _ in _); [exact B | | ];
    [typeclasses eauto|]
  );
  unshelve refine (let z := _ in _); [exact C | | ];
  [typeclasses eauto|];
  let av := eval unfold x in x in
  let bv := eval unfold y in y in
  let cv := eval unfold z in z in
  refine (Build_X av.(@a) bv.(@b) cv.(@c));
  change X
.

Ltac list_instance A :=
  try (
    let x := fresh in
    unshelve refine (let x := _ in _); [exact A | | ];
    [typeclasses eauto|];
    let ans := eval unfold x in x in
    clear x;
    idtac ans;
    fail
  ).

Hint Extern 2 X => instance : typeclass_instances.

Goal True.
list_instance A.
list_instance B.
list_instance C.
list_instance X.
```

CohenCyril:
The code you are giving is a bit too ad-hoc for me, so I'm trying to build reusable components instead, like:
```coq
Class cut (class : Type) := Cut {cut_field : class}.
Notation "'cut c" := (@cut c) (at level 200).
Hint Extern 0 (@cut _) =>
  eapply @Cut; repeat (eapply @pair || eapply existT);
  once typeclasses eauto : typeclass_instances.
```

ppedrot:
@CohenCyril Yes, but that's precisely why it'll fail. You're facing the expressivity wall of Ltac here. My encoding is pretty much systematic though. The `instance` tactic is easily accommodated to handle an arbitrary number of superclasses, cuts and an arbitrary body.

maximedenes:
@ppedrot Why do you say it will fail? It seems to work fine.

ppedrot:
I mean, managing to write a set of combinators to encode arbitrary cuts.

mattam82:
@ppedrot Precisely. I wondered if with exceptions we could emulate this "long" once, but the hard part seems to be figuring out where the start of the cut is.

text/typeclass_control.md
mattam82: @CohenCyril Yes, interleaved searches are a mess indeed, and a few times I've wondered how we could get more control in Hint Extern to control the backtracking of typeclasses. Maybe with Ltac2 we could get a handle on the continuations involved to express more elaborate control.

CohenCyril:
@mattam82 you are right, once does not implement cuts fully, because prolog cuts should also never consider then next rule/instance when the solution of the arguments before the cut are found...

ppedrot:
@CohenCyril I don't understand what you're saying. Do you have a Prolog example whose semantics is not the same as the Ltac one I gave above, and its intended semantics?

mattam82:
@CohenCyril yep, that's another difference. It should be possible to implement globally in the search algorithm though. I wonder if you're adamant about having the "cut" in the type or if a separate [Hint Cut instance after 2 4 : typeclass_instances] would work as well?

CohenCyril:
@ppedrot here is the prolog example (from https://en.wikibooks.org/wiki/Prolog/Cuts_and_Negation and slightly modified)
```prolog
a(X) :- b(X), !, c(X).
a(X) :- d(X).

b(1).
b(3).

c(3).

d(4).
```
Which fails when asked `?- a(X).`

But it succeeds when translated to Coq using my encoding:
```coq
Class cut (class : Type) := Cut {cut_field : class}.
Notation "'cut c" := (@cut c) (at level 200).
Hint Extern 0 (@cut _) =>
  eapply @Cut; repeat (eapply @pair || eapply @existT);
  once typeclasses eauto : typeclass_instances.

Class a (x : nat).
Class b (x : nat).
Class c (x : nat).
Class d (x : nat).

Instance a_bc x `{'cut b x} `{c x} : a x.
Instance a_d x `{d x} : a x.
Instance b1 : b 1.
Instance b3 : b 3.
Instance c3 : c 4.
Instance d4 : d 4.

Goal exists x, a x.
Proof.
eexists.
Set Typeclasses Debug.
typeclasses eauto.
Qed.
```
(Of course, removing the `a(X) :- d(X).` rule, the behavior is correct and the resolution fails in both prolog and Coq.)

@mattam82 I would be happy with a separate Hint Cut instruction, but maybe with named arguments instead of numbers, like `Hint Cut instance x ! _ _ ! _ z.` (a bit like the `Arguments command` to decide when to simplify).

mattam82:
@CohenCyril that sounds right.

CohenCyril:
@mattam82 after discussing with Maxime and Enrico, it would seem appropriate to go through an other kind of Hint, `Hint IfExtern` taking two tactics `cond` and `then` as arguments that `eauto` would combine using `IFCATCH cond then other_hints` instead of `ORELSE`.

mattam82:
@CohenCyril Would that be enough to express cuts then?

CohenCyril:
@mattam82 cuts can be defined using` IFCATCH (ONCE cond) th el`, so at the Hint Extern level, it would indeed be enough. However, it would still be desirable to be able to annotate instances using cuts, which would translate to the appropriate combination of hints and once.

CohenCyril:
@mattam82 And there still is the problem that relaunching` typeclasses eauto` starts a whole new search (losing potential options passed to the tactic, starting the trace from 1. again and interleaving the printing)...

mattam82:
We implemented a first prototype here: https://github.com/mattam82/coq/tree/hint_extern_if

text/typeclass_control.md
CohenCyril: @mattam82 any progress on that?

PR #32, opened 2017-10-16
RFC: Disentangling erasure and proof irrelevance

aa755:


gares:
Thanks for your RFC.  CC @letouzey  and @SkySkimmer

mattam82:
Hi Abhishek, this seems to go against models of CIC with a quote operator as in 
https://www.pédrot.fr/articles/cpp2017.pdf

ppedrot:
@mattam82 The impredicativity of Prop already breaks it actually. More generally, the majority of syntactic models I know are breaking singleton elimination, except for reader-like effects like forcing. Even parametricity cannot interpret an erasable Prop properly, see [this article](https://hal.archives-ouvertes.fr/hal-00730913/document). So there is something fundamentally broken with the idea of extraction erasure, or put it more positively it makes a very strong assumption on the computational model of the corresponding theory.

That is not really a concern, it simply means that one cannot use `ESet` in those extended models of CIC. But it also probably means that the construction is too specific for its own good. I find the solution proposed in the article by Keller and Lasson mentioned above to be already one step higher on the ladder of generality.

mattam82:
I see. Isn't ``ESet`` actually a kind of synonym to Keller & Lasson's ``Set_i`` here?

ppedrot:
@mattam82 Morally yes I guess, but the CEP doesn't provide any kind of typing and subtyping rules, so it's hard to judge in the details.

aa755:
Thanks all for the feedback.

I don't see the analogy with Keller and Lasson's paper.
They say "In the refinement (CIC_r), the CIC hierarchy of sorts Type is thus divided into two classes: a
hierarchy of *sorts Set, whose inhabitants have a computational content*, and a hierarchy
of *sorts Type, whose inhabitants are uninformative*.". 
My design is actually the opposite, on both those issues:
First, I am proposing that the types in `ESet` be (computationally) uninformative (by restricting pattern matching). 
Second, I am not proposing that the inhabitants of `Type` be uninformative. 
For example, they disallow inductives in `Type` (see their rule 2 in their Sec. 3.1), perhaps to ensure the uninformativeness of types in `Type`. 
I don't disallow it. I am just adding a new universe, `ESet`, which is dedicated to erasure, and not changing anything about how the existing universes work in Coq's typesystem.

It seems that the generality concern raised by @ppedrot comes from the fact that there is only one `ESet` and not a chain. If the chain is really needed, we can have `ESet_i`. Unlike in CIC_r (the refinement in Keller and Lasson), where disallowing to `Set_i:Set_{i+1}` (thus avoiding the need for `Prop:Prop`) was fundamental to the problem they wanted to solve, I see no problem with having `ESet_i:ESet_{i+1}`. Of course, then `ESet_i` would be translated to `Type_i` in the consistency proof, exactly as Keller and Lasson do.

I agree with the criticism that I don't formally provide the typing and subtyping rules for `ESet`. I'll try to work them out. Is it necessary to add subtyping rules? Is it fine for  `ESet` to not have any subtyping relationship with any other universe?

PR #33, opened 2018-02-23
Goal with state

gares:


ejgallego:
Well, the main question is that given that tactics are a monad, with the natural parameter `'a` being a per-goal parameter, it is not so clear why you cannot use that parameter. So far we have that objects of type `unit tactic` are "of tactics producing goals with no other attached information".

> it is unclear which static benefit one would get.

So tactics that would be, for example, of type `to_clear tactic`, would be marked a "hey, this goal has produced some to clear info! The consumer of the goal must either pass it, or take care of it somehow". 

My dilemma comes from the following reflection: on one hand we have a type of tactics, that is a monad, thus can output things. On the other hand, it is claimed that this doesn't work, and all the operations are of type `unit tactic`. Now, it turns out that what we are proposing to have two mechanisms very similar in spirit. This sounds highly suspicious. Then, either one of them is superfluous, and it should be then removed, or either both are broken and a new one should appear.

maximedenes:
> with the natural parameter 'a being a per-goal parameter

That's not the meaning of the parameter in the current engine IIUC.

ejgallego:
> That's not the meaning of the parameter in the current engine IIUC.

What it is then?

`tactic` is a pretty standard "bag" monad with failure, so in this case each `'a` is associated to an open goal as far as I can see.

maximedenes:
Well, an `'a tactic` applies to several goals at once, so I'm not sure what you mean by a "per-goal parameter".

maximedenes:
(I find it would be a reasonable semantics btw, just saying it's not the current one IIUC)

ejgallego:
> Well, an 'a tactic applies to several goals at once, so I'm not sure what you mean by a "per-goal parameter".

I mean that every goal has now attached a value of type `'a`. What applies to several goals at once is the type. Then we have that given `fmap : ('a -> 'b) -> 'a tactic -> 'b tactic`, in `fmap f t`, `f` will be applied as many times as open goals we have [modulo laziness but that's a different story].

ejgallego:
So this state proposal seems suspicious because it overlaps with the natural parameter of `tactic`. On the other, the `tactic` itself seems quite suspicious due to the existance of `enter`.

Isn't it funny that what have so much boilerplate of the form `enter (fun gl -> let g = goal x...`. I am convinced it is not a coincidence.

maximedenes:
Again, I don't understand. For example, `cycle` has type `'a tactic`. What does it do on its parameter? It is the identity. What does it do on goals? It permutes them. So if you chain `tac1 ; cycle n; tac2`, I don't understand how you maintain this view of the parameter being the type of values attached to goals. But maybe I simply don't understand what you are saying.

maximedenes:
> Isn't it funny that what have so much boilerplate of the form enter (fun gl -> let g = goal x...)? I am convinced it is not a coincidence.

And I fully agree with that. The current postulate that is made by some developers is that the "real" notion of tactic is something that applies to several goals and returns several goals, whereas in practice all of them except `eauto` and `cycle` apply on *one* goal and return several. The algebraic structure that resulted from it was not pleasant, or not well known enough, so they were embedded into something that is more standard but denies truth (with slogans like "there is no `;` in `; [ ... | ... | ...]`). But this entire part of the discussion is orthogonal to this PR and should happen somewhere else.

ejgallego:
`cycle` does nothing to the parameter, as `tactic` is a monad the parameter is natural and it is attached in this case to each goal.

If you have goals `[(g1,p1);(g2,p2)]` then `cycle` will transform that into `[(g2,p2);(g1;p1)]`.

I am not maintaining any view, what I say it is just a consequence of `tactic` being monad.

maximedenes:
> If you have goals [(g1,p1);(g2,p2)] then cycle will transform that into [(g2,p2);(g1;p1)].

Where are such pairs of goals and data implemented today?

ejgallego:
> But this entire part of the discussion is orthogonal to this PR and should happen somewhere else.

I am not sure I can agree with that. This CEP is proposing a modification to the inner core of the tactic engine, so I stand on my view that we shouldn't follow the approach that two wrongs make a right.

If we need to reify the engine as to have different invariants, we shall then prepare a roadmap and fix it, not hide our head in a hole-hack.

ejgallego:
> Where are such pairs of goals and data implemented today?

Hehe, it takes a while to peel the onion to get at:
```ocaml
  type ('a, 'i, 'o, 'e) t =
      { iolist : 'r. 'i -> ('e -> 'r NonLogical.t) ->
                     ('a -> 'o -> ('e -> 'r NonLogical.t) -> 'r NonLogical.t) ->
                     'r NonLogical.t }
```

ejgallego:
In a monad, what you really care is the type of `bind` (or `>>=`). In this case, even if the comments claim that:

> Tactics are objects which apply a transformation to all the subgoals of the current view at the same time

`bind` still works at a single subgoal level, unless I am extremely wrong.

maximedenes:
> bind still works at a single subgoal level, unless I am extremely wrong.

I believe that's not correct.

ejgallego:
> I believe that's not correct.

Let's have a look at `plus`.

ejgallego:
But the comment says:
```      In that vision, [bind] is simply [concat_map] (though the cps
      version is significantly simpler), [plus] is concatenation, and
      [split] is pattern-matching. *)
```
and I'd believe it. Too lazy to get pen&paper now.

ejgallego:
`plus` is concatenation, so `plus (return 1) (return 2) == [1;2]`.

ejgallego:
So in the semantics you propose, what would the semantics of `plus (return 1) (return 2) : int M.t`

ejgallego:
Well, I realize that I am using goal here totally wrong. I am using goal with the meaning of a goal **in proof search** whereas Maxime is using goal in the "Coq" sense, so I think Maxime is correct.

But we are back to the question of why Proof-Search goals and Coq goals do differ indeed.

SkySkimmer:
What is a proof search goal?

ejgallego:
Something that either succeeds or fails, with a disjunctive structure, whereas in the Coq sense, goals a conjunctive in the sense that if you have N goals you need to solve all of them.

maximedenes:
> I am using goal with the meaning of a goal in proof search whereas Maxime is using goal in the "Coq" sense, so I think Maxime is correct.

Exactly, that's the difference I tried to make (I wrote "successes" on another channel for that). Sorry I'm not very clear, I'm taking care of my son at the same time ;)

gares:
It is OFF TOPIC, but since I spent a month writing code in this API I think I've understood the intuition of the authors. In case it was not clear, I don't like this API, I'm just explaining it.

The `'a` is the monad is somewhat used to impose the following discipline:
- all combinators are written with `'a = unit`, so `>>=` is useless, hence its specialization `<*>`.
- even `Goal.enter` that all tactics have to call in order to do anything sensible (e.g. look at the goal) has `'a = unit`. Note that when you `Goal.enter f`, `f` is called on all goal (in order, and the order matters, since there is a state, the evar_map, hence some comments above are just false)
- the only brick that lets you use the `'a` is `Goal.enter_one` that asserts there is only one goal, and lets you `return data` (hem, `tclUNIT data`). In practice if you want to generate data using the goal you have to `Goal.enter (fun g -> ... Goal.enter_one f >>= fun data -> ..)`.

The intuition is that the `'a` is for goal-related data. *But* since "tactics" work on a bunch of goal, and not just one, one would need to put `'a list` (instead of `unit`) everywhere, and then the programmer would need to handle this list of goal-related data appropriately, e.g. cycle would shuffle the goals (in AND),  who would shuffle this list?

So, all "muti goals" combinators are flagged as `unit`, when you focus on one goal you finally get to use the `'a`, and you lose such freedom as soon as you call another combinator (with a few exceptions, requiring the user to provide as many tactics as goals, and so on).

At least, this is the intuition I got using this API: `'a` is not the `'a T` in a monad `T`, it is a flag make `bind` unusable in the multi goal setting (you can use it only in the mono goal setting).

If you add to that the observation that 99% of tactics is mono goal, you understand why I don't like all this infrastructure. 

Going back to this CEP, the PR is *not* a way to improve this API. As I say in the CEP, I had this code also before the tactic monad. One has still to use `enter_one` in order to use `>>=`, etc. This facility lets one attach data to goals (as in Coq goals), nothing more, nothing less

ejgallego:
Sorry Enrico but this is not OFF TOPIC at all, this is a CEP modifying the **core** proof datatype of the engine so of course discussing the current state of the engine, and trying to consider alternate solutions is 200% on topic.

Please, let's leave this kind of "confrontational" dynamic behind us, we should all be happy to get our code questioned and criticized. It seems to me that many problems in the current codebase are a result of people getting burnt out and dropping unfinished code in as a kind of surrender act. We need to do better than that. If the core engine needs modifications, we should instead go and tweak in place of adding more stuff on top of shallow foundations.

As a said before, IMO it is OK to merge any kind or PR even if it is the ultimate dirty über hack, as long as there is a clean understanding of the situation and of what the roadmap/future steps are.

Regarding this PR/CEP, I think it looks OK, however it is basically isomorphic to the current `evar_info`, and while I agree that can be useful in the short term, it seems to me that it suffers from the same problems the `evar` case has and I wouldn't recommend this style of programming to tactics writers. In a large codebase, using _untyped_ arbitrary state that is not reflected in types is a sure recipe for a maintainability hell. Imagine I am a tactic writer, and I am composing with tactics using the store. Like adding a new option, adding a new mechanism tends to increase maintainability cost geometrically. **There is no guarantee that I won't drop/mess with your state**. I could manipulate the `comb` in many ways and most of them are going to be bad for the store. Anyways, I guess the real question here is if the PRs that need this feature should go in, which so far is not clear if I understand the discussion there correctly.

Now going back to the discussions about the foundations, unfortunately I think that the monad is indeed not very useful as of today for many use cases, including yours. However I think that there is an unanswered question here, why are all combinators specialized to `unit tactic`? Is that a fundamental limitation? What is so special about `enter` that cannot have a more general type?

gares:
> However I think that there is an unanswered question here, why are all combinators specialized to unit tactic? Is that a fundamental limitation? What is so special about enter that cannot have a more general type?

I think it could, but would be `(goal -> 'a tactic) -> 'a list tactic`.
Even so, it would not solve the problem this CEP  tackles, since the list of `'a` would not be attached to the goals, it would be up to the tactic writer to keep such invariant. The best example of the problem is the one I call HARD2 in the CEP: `case t >>= fun names -> try done <*> block_intro names`. The names generated by case are "1 list per goal" but `block_intro` is called on less goals (since `try done` may have killed some) and hence `names` would be too long. This cannot be solved by given to `case` (and `Goal.enter`) the more general type above, nor by making the data that flows explicitly types, since the property I want is purely dynamic.

ejgallego:
> I think it could, but would be (goal -> 'a tactic) -> 'a list tactic.

Well, not even that will work. The monad only knows about backtracking, the refinement structure (usually a tree+option monad) is just not there to take advantage of. As you point out, it would be up to the tactic write to maintain the proper `comb`, but this is already what basically happens today.

> HARD2 in the CEP: `case t >>= fun names -> try done <*> block_intro names.`

Umm, that looks like it should be: `case t >>= try done >>= fun names -> block_intro names` where indeed `try done` is natural. But yeah, the structure for the refinement/conjunctive part of the engine is just not there so short term the best that can be done is messing with "`comb`" and you propose.

I am getting convinced that many of these problems do indeed arise due to the lack of the conjunctive solving operation [a super-primitive `refine`] as I tried to raise in https://github.com/coq/coq/issues/6608 , but indeed that's not so easy to do.

herbelin:
@gares: As far as I understand, whether it is `_` or `+`, these are patterns which require to maintain a state, in order to collect actions to be done only at the end of the interpretation of a list of patterns. I had yesterday a look at `ssripats.ml` and, apparently, the `ssr` code maintains this state in the `tac_ctx` record (field `delayed_clears`), which is eventually taken into account in `with_fresh_ctx`. You also have a variant `tac_ctx tac_a` of `V82.tactic` which basically is equivalent to `tac_ctx -> tac_ctx * V82.tactic`. So, you basically have an interpretation function `ipattac` of type `ssripat -> tac_ctx -> tac_ctx * V82.tactic` with a continuation `with_fresh_ctx` of type `(tac_ctx -> tac_ctx * V82.tactic) -> V82.tactic`. So, why don't you simply replace `V82.tactic` with `unit tactic` (as is done in Vanilla `Tactics.intro_patterns_core` actually)? As far as I can judge, this would make your life simple without needing the changes of the engine that you are asking for. Or do I miss something?

gares:
> Or do I miss something?

You are looking at the code that turns a non-ctx tactic into one that handles the context (by just preserving it).

What is relevant in the code you are looking at is that `'a tac_a = (goal * 'a) sigma -> (goal * 'a) list sigma`. This is exactly what the PR does.  Since the `goal` and `goal list` are not part of the monad (the `comb`) this is where I've added the `'a` (actually an opaque `state`, hence some comments by Emilio).

herbelin:
See my [comment](https://github.com/coq/coq/pull/6700#issuecomment-368437484). The new implementation is the result of a big effort and that's great. I was just saying that the `'a` does not have to mandatorily be in the "Proofview.monad", though it could, as you did, or suspectingly alternatively as another field of the record `Proofview.proofview`, but it is ok as you did. Congratulations for the work.

maximedenes:
> as another field of the record Proofview.proofview

That's a good point. IIRC, I had considered this when implementing, not sure what the conclusion was.

aspiwack:
This all happened as I was not looking, I'm sorry.

I must say that there are a number of use-cases which need some state-bound goals like this. (mostly I found this need when considering how some proof search strategies could be implemented, I don't remember the details).

So I'm happy that solutions are being proposed here.

My only worry, with this solution, is that every tactic must specify their behaviour with respect to goal states. This seems to be missing here, unless I missed something? I think this CEP should be augmented with a detailed explanation of what happens to state in all the primitive tactics (I'm guessing that most tactics don't have much of a choice, what about `refine`?).

Also a link to the PR corresponding to the implementation of this CEP.

PS: @gares your characterisation of the tactic monad API is essentially entirely wrong. I don't know where you got such an impression.

ejgallego:
> I'm guessing that most tactics don't have much of a choice, what about refine?).

We asked somewhere about what `refine` and indeed there is code taking care of the state [I think it is propagated to all subgoals if any, but I don't exactly recall now]

> your characterisation of the tactic monad API is essentially entirely wrong. I don't know where you got such an impression.

A lot have been said about this PR/CEP and discussion is unfortunately a bit sparse, so it is hard for us to track which comments you are referring to, would you mind being a bit more concrete?

ejgallego:
> I just believe it ought to be part of this CEP

+1 

> Sorry: this one.

Umm, well indeed the way the conjunctive structure of the proof obligations is handled has confused me for a long time. So in principle, there is no other mechanism than either `enter` which indeed does suffer from the shortcoming Enrico points out, or by relying on the low-level `tcl*GOALS`, which seems to expose too much detail.

So how's the conjunctive part supposed to work?

gares:
Arnaud I guess you talk about the unit argument? Explain the correct view then plz

aspiwack:
Sure.

The `'a` in `'a tactic` is the `'a` of a monad. Because `'a tactic` is a monad. The purpose of having a monad is to be able to abstract over the tactic implementation, while giving a compositional semantics. Especially with the perspective of compiling a programming language to the abstract interface.

Speaking of programming languages, you might have noticed that Ltac has a lot of quirky behaviours. A number of these, like not being able to write an expression that modifies the goal _and_ return a value, come down to the fact that we didn't have a tactic monad at the time.

Anyway, the issue here is that the `'a` cannot be attached to individual goals: it wouldn't be associative. I explain this reasoning in [Chapter 4 of my thesis](http://assert-false.net/arnaud/papers/thesis.spiwack.pdf#chapter.4). Though the details of the design has changed since then.

This is annoying for some application (like many proof search strategies) and beneficial for others (you can use the tactic monad without a goal, for instance after you close a proof, you could print an information message. A usage which has seen some success is the `numgoals` tactic).

I will agree with you that the `enter_goal` API is not entirely satisfying. Certainly passing a list around to match the goals is not in general a very good idea. You can enter single goals and extract values (which means that the tactic `1: return x (* x an hypothesis of goal #1 *)` is legal). It's useful. But for your use-case, this is not sufficient.

Does that shed a bit of light? (cc @maximedenes who asked me similar questions recently)

maximedenes:
Something I never had the time to check is how does Lean's tact monad compare to yours. Can you say a bit about that?

aspiwack:
I frankly have no idea. I'd love to know. But likewise: time and all this.

@mattam82 you've looked at Lean before, if I'm not mistaken. Do you know something about their tactic monad?

Zimmi48:
@aspiwack You say that the `'a` is not sufficient to implement many proof-search strategies: it seems to be though to emulate `eauto` behavior using a new `tclFOLD` tactical (cf. https://github.com/coq/coq/pull/721/commits/35e1ac702c58ffd7c463cb0f982b6f64ab0cae61 whose initial author is @cmangin). We have not explored how the same could have been done with the new goals with state (the initial commit by @cmangin was done much before, on June 1st, 2017).

aspiwack:
@Zimmi48 Yes, `eauto` is the ideal use-case, in a sense. But other proof search strategies may want to overlay each goal with some structure specific to the algorithm. And that's already not obvious, and the sort of things that @gares is trying to address here.

gares:
Can you explain why you put `unit` (almost) everywhere? The reason for this is what I was trying to guess, and to my understanding of monads, a unit monad is not very common.

gares:
I missed one ingredient: *all* tactics (as in tactics the user calls, not internal code) but for 2 (cycle and eauto) do need to access the current goal, hence they are flagged as `unit`. While the monad may stay a monad (i.e. keep associativity) it is also a very weird one to me, and the feature of being able to pass data to the following tactic is kind of lost. What do I miss?

aspiwack:
The answer will be disappointing: it's because all the tactics written prior to the new engine were all this form, as this was the only possible thing to express.

If you'll allow me an analogy: suppose you are porting a Fortran code-base to C, then someone may be wondering why almost all your functions return `void`, and if most functions output their result by mutating an argument, why do we need to be able to return values?

This is a bit the same situation.

I'll gladly admit that we are still looking for the best ways to leverage the monadic structure. Beyond Ltac2 which uses it for its semantics. And yes, the `Goal.enter` situation is not particularly satisfactory. I'd love to know a better way.

ejgallego:
IIUC so far the monads deals with the disjunctive fragment of the proof search. But have the conjunctive part? How are we supposed to handle it?

What I inferred so far is that the tactic engine provides no support for this case, other than the low-level, imperative/Unsafe functions. Is that correct?

mattam82:
AFAIR it’s the classical goal * state -> goal list * state interface, but I
should check the ICFP paper again. About passing of state etc I think we
should give a closer look at Dependent LCF of Harper et al which builds on
the same foundation as Arnaud’s monad. It might have the programming
structures we’re looking for.
Le sam. 31 mars 2018 à 06:26, Arnaud Spiwack <notifications@github.com> a
écrit :

gares:
The disjunctive part is hidden in the monad that gives you backtracking.
The conjunctive part is hidden behind `Goal.enter` that gives you access to each and every goal (one at a time).
Whenever you use `Goal.enter` the `'a` parameter of the monad becomes `unit`, making "bind" just about sequencing.

@mattam82 that type is only there for V82 tactics.
BTW, I tried to read the paper but I could not get what it does better than what we do today.

ejgallego:
> The conjunctive part is hidden behind Goal.enter that gives you access to each and every goal (one at a time).

Umm, IMHO `Goal.enter` doesn't provide access to the conjuctive part. In particular, there is no operation to solve a goal or to create new conjunctive ones.

ejgallego:
This lack of interface is somehow solved by the hack `refine` is today. In a sense, you'd like to split refine in two, one part which is the tree refinement algebra [a substitution monad if you want] and the high-level refine that calls the pretyper etc...

But yeah, talking is very easy, doing the code is another story.

gares:
I'm not very familiar with the jargon, but to me `refine` does two things: calling the pretyper on the term and collecting residual holes that become new conjunctive goals. 

It is essentially what "backchain" does in prolog: unify the head (here means typecheck t and unify the resulting type with the goal) + generate a new goal for each premise (here a new goal for each unif variable still unresolved in t).

Is this what you mean?

ejgallego:
> I'm not very familiar with the jargon, but to me refine does two things: calling the pretyper on the term and collecting residual holes that become new conjunctive goals.
> 
> It is essentially what "backchain" does in prolog: unify the head (here means typecheck t and unify the resulting type with the goal) + generate a new goal for each premise (here a new goal for each unif variable still unresolved in t).
> 
> Is this what you mean?

Not exactly, unification in Prolog is never gonna create more conjuctive goals which are usually fixed by the clause compilation.

Here the setting is a bit different, your basic conjuctive operation is "substitute", which may create more obligations. This forms a monad on the type of the holes with `return x = Var x`, and `bind` the substitution. Then the pretyper / refine should use this API to indeed refine / create new goals, however this is not what happens as the conjuctive part of the engine has no structure AFAICS.

This is a problem as then basically the API guarantees nothing. [You can have leaks, out of sync goals/holes, etc...]

gares:
Thanks now I see what you mean.

I'm afraid some properties are too dynamic to be assured by any API.
The fact that the logic mixes proof evidence (terms) and statements (types begin unified) makes it hard, for example, to ensure that unification does not solve a goal. You can force all goals for which the (missing) proof evidence occurs in a statement as non-goals (shelved goals), but this is not always what you want in practice (eg people ask for unshelve & co).

IMHO the evar_map is a heap (as in prolog's heap). Some entries in the heap are used to represent goals (they are essentially pairs, a non assigned variable representing the proof term and its type).
Tactics take a heap and return a heap. They also take some arguments (that may point to heap cells) among which a goal. They return an updated heap, and a list of what should be considered a sub goal, i.e. heap cells the caller should eventually pass to other tactics (as their goal argument). It is not easy to statically know which part of the heap (which goals) a tactic can alter, since an input term may point to a goal (or not). Hence dynamically one has to check if the goal is still open (today done by Goal.enter).

This is crystal clear in the V82 API where tactics are `goal * heap -> goal list * heap`. The type does not compose, hence you need tacticals (a scheduler) to combine tactics. The semantical trick is that a tactic called on an already solved goal behaves like `idtac` (even if it is `fail`).
In the current API the `goal list` is sweeped under the rug (it is the future_goal thing). Why is it hidden? The answer is, I'm afraid, to pretend things compose "well".

gares:
BTW, this is the paper @mattam82 was suggesting: https://arxiv.org/pdf/1703.05215.pdf While I lack the mathematical background to understand it fully, you probably have.

aspiwack:
@ejgallego I think it is better to start by reading [Chapter 4 of my thesis](http://assert-false.net/arnaud/papers/thesis.spiwack.pdf#chapter.4). Details have changed, but most of the reasoning is already there.

PR #34, opened 2018-03-26
RFC: Representation of return predicate and branches of "match" in the implementation, and how reduction tactics should treat them.

herbelin:
Rendered [here](https://github.com/herbelin/ceps/blob/reduction-in-inductive-predicate-and-branches/text/inductive-branch-predicate-representation-and-reduction.md).

### DIFF
--- a/text/inductive-branch-predicate-representation-and-reduction.md
+++ b/text/inductive-branch-predicate-representation-and-reduction.md
@@ -4,6 +4,10 @@
 
 ----
 
+## Summary and motivation
+
+Regarding the branches and the return clause of a `match`, there is a discrepancy between the user-level syntax and the internal representation. This has unexpected effects on how some reduction tactics consider a `match`. This CEP is about proposing a canonical internal representation which matches the user view, as well as about clarifying the expected behavior of reduction tactics on `match`.
+
 ## Context
 
 The user view at inductive types is that of an `as x in I realargs return P` for the return predicate and of a `C vars => u` for a branch (view 1). This view matches the "pure" (dependent-product independent) proof-theoretic view at inductive types. But the implementation represents the return predicate and the branches as functions (view 2).
--- a/text/inductive-branch-predicate-representation-and-reduction.md
+++ b/text/inductive-branch-predicate-representation-and-reduction.md
@@ -1,0 +1,126 @@
+- Title: Representation of predicate and branches of inductive types in the implementation, and how reduction tactics should treat them
+- Driver: Hugo
+- Status: Draft
+
+----
+
+## Context
+
+The user view at inductive types is that of an `as x in I realargs return P` for the return predicate and of a `C vars => u` for a branch (view 1). This view matches the "pure" (dependent-product independent) proof-theoretic view at inductive types. But the implementation represents the return predicate and the branches as functions (view 2).
+
+The advantage of representing the return predicate and the branches as functions is that there is no special treatment to do in `match` (internally `Case`) to deal with the binders of the return predicate and branches: they reuse the path code of `fun` (internally `Lambda`).
+
+The drawbacks are that:
+- no insurance that the predicate and branches have the η-expanded form expected for printing; no insurance to have enough abstracted names and necessity to invent names
+- the notion of subterm in the internal representation does not match the notion of subterm in the user representation; a whole branch is itself a subterm; occurrences of a subterm are unstable by invisible internal η-conversion
+
+To be more precise, from one part of the code to another, the treatment of branches and return predicate can be more or less ad hoc, ranging from view 1 to view 2 with variants. For instance, `vm_compute` and `native_compute` considers the branches and return predicate under view 1, but the other reduction machines (`cbn`, `cbv`, `lazy`) consider them under view 2. Tactic `pattern` considers them under view `2`. Also, the code for typing branches in the kernel now assumes view 1 and may fail in interaction with parts of the code which assume view 2 (see #6749, #7068, and probably other reports).
+
+## Conceptual differences in term of typing rules
+
+### 
+
+The typing rule for `match` implicit in the user syntax is:
+```
+Γ, Δ[params/Ω], x : I params |Δ| ⊢ P : s
+Γ, Ξᵢ[params/Ω] ⊢ ui : P[realargsᵢ/Δ][Cᵢ |Ξᵢ|/x]    (for all i < nb constructors)
+Γ ⊢ c : I params realargs
+---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gares:
The CEP template file proposes a model with sections (in a specific order).
Why didn't you follow it?

herbelin:
The first reason is that I did not remember that there was a model. What would you like me to change in what I did? Would you like another order of sections?

herbelin:
@gares: I added a "TD;LR" section, assuming that this is what you felt was missing.

maximedenes:
> The CEP template file proposes a model with sections (in a specific order).
> Why didn't you follow it?

@gares is that a PR template? If not, it should be turned into one.

mattam82:
I like the idea of having a canonical representation. For example, CertiCoq has a phase to go into the (Δ, brs) style for pattern matching branches, avoiding dummy beta-reductions at run-time, but tha phase creates dummy β-η redexes to be removed later by another phase, which is a pity (but it's not too bothersome either). I just had not anticipated this problem because the encoding of branches was not clear to me at first (always η-expanded or not in particular). My only worry is that much code does its own Array.map foo brs on branches currently, but maybe I'm overestimating that. I think the part about let is a bit orthogonal here, no?

herbelin:
> My only worry is that much code does its own Array.map foo brs on branches currently, but maybe I'm overestimating that. 

I already started to develop a library of combinators to replace these `Array.map` by some kind-of `map_branches`. So, I could almost propose something in the next few days.

> I think the part about let is a bit orthogonal here, no?

Yes and no. Yes, in that we can adopt a canonical representation without the lets. No, in that we have to make a decision wrt lets in adopting a canonical representation. Or maybe I don't get what you mean by "let" is orthogonal.

gares:
> @gares is that a PR template? If not, it should be turned into one.

At the root of the repo there are 2 file, a README and a file called 000-template that is also mentioned in CEP0... If you know how to, please turn it into a template.

In any case my point was : please make the motivation clear. I think @herbelin added this section in his second commit.

Zimmi48:
It can't really been turned into a PR template because PR template are for the PR description, not the PR content.

herbelin:
Hi, I'm progressively implementing this ceps in a private branch. The current state is the following:
- `lazy`, `cbv`, `cbn` now see branches of a `match` as part of the structure of a `match`
  - in particular, binding arguments of the constructor to the pattern variables is now pure `match` reduction and does not require the  `beta` flag anymore (for instance `Eval lazy match in match 2 with 2 => 1 | _ => 0 end` now reduces to `1` without needing `beta`)
  - similarly,  `lazy`, `cbv`, `cbn` do not contract anymore the `let`s which are parts of the signature of a constructor (see coq/coq#7068)
- I started to make branch and predicate combinators for `map` but not yet for `fold` and `iter` (see coq/coq#7109)
- I introduced a kernel check that the internal structure of the branches of a `Case` follows the `Lambda`/`LetIn` structure of the signature of the type of a constructor (see discussions coq/coq#6749 and coq/coq#7068)
- Preparing to the possibility to use `LocalAssum`/`LocalDef` rather than `Lambda`/`LetIn` in the structure of branches of a `Case` (see coq/coq#7080)

Any new comments about continuing in this direction?

ppedrot:
Although I already said that somewhere else, let me repeat here that I am strongly supporting this CEP. What is the expected schedule?

herbelin:
> What is the expected schedule?

I have a version which provides combinators for an abstract version of branches and return clause. This also includes fixes for reductions in `Reductionops`, `Cbv`, `CClosure`. I have a little regression on the proofs produced by `case`/`destruct` (they artificially introduce beta-expansion).

I should probably cut into pieces easier to review: it is currently 19 commits.

So, the schedule is probably this winter.

ppedrot:
FTR, out of procrastination I started working on a direct implementation of the second change, i.e.
```
type ('constr, 'types) pcase_branch =
  ('constr, 'types) Context.Rel.pt * 'constr

type ('constr, 'types) pcase_return =
  ('constr, 'types) Context.Rel.pt * 'types

type ('constr, 'types, 'sort, 'univs) kind_of_term =
  ...
  | Case      of case_info * ('constr, 'types) pcase_return * 'constr * ('constr, 'types) pcase_branch array
  ...
```

I don't know how far @herbelin went, but my take on that is that this second step should just be a temporary one. I suspect that erasing the types from the binders (which are inferable) would yield important efficiency gains, because they account for a important part of the term size. In particular, hashconsing and evar-normalization would probably benefit a lot, and maybe reduction as well. More precisely, I think we should perform the following change:

```
type ('constr, 'types) pcase_branch =
  ('constr, unit) Context.Rel.pt * 'constr

type ('constr, 'types) pcase_return =
  ('constr, unit) Context.Rel.pt * 'types

type 'constr pcase_param = 'constr array

type ('constr, 'types, 'sort, 'univs) kind_of_term =
  ...
  | Case      of case_info * ('constr, 'types) pcase_return * 'constr pcase_param * 'constr * ('constr, 'types) pcase_branch array
  ...
```
where `pcase_param` is the array of parameter of the term being destructed. As of today this information is present in the return clause, but with this encoding scheme recovering it without setting it in the case elimination would require full-blown typing, which might be too costly.

I haven't tried implementing this yet, but it's probably going to be easier after having the second part of this CEP done.

SkySkimmer:
>`('constr, unit) Context.Rel.pt`

Why not (unit, unit)?

ppedrot:
@SkySkimmer You still need the let bodies to compute kernel reduction without calling the environment.

herbelin:
@ppedrot: What I implemented so far is [here](https://github.com/herbelin/github-coq/tree/abstracting-cases). Part of it is ready but with a little regression on the proofs produced by destruction tactics. They introduce useless beta-expansion in the branchse and and I wanted to fix that.

ppedrot:
@herbelin OK, so it seems your set of patches is quite orthogonal to mine. You're dealing with the upper layer changes of behaviour, while I'm just trying to change the kernel representation. If you're not planning on changing the kernel representation, I can rebase mine over yours, if that's not too costly.

herbelin:
> You're dealing with the upper layer changes of behaviour, while I'm just trying to change the kernel representation.

I was progressively abstracting over the type of branches (hence the `type 'a branch = 'a`) so as to eventually be able to change the representation. My methodology was to make `branch` abstract, to make some progress at changing the code where it was relying on the representation, and to remake it concrete so that it compiles with the part of the code which I did not have adapted yet.

> If you're not planning on changing the kernel representation, I can rebase mine over yours, if that's not too costly.

I don't have time and thus concrete plans on this branch before (a minima) February. If you basically agree with the changes already done, feel free to rebase on it and extend it. (I can also do the rebase of the branch over master myself by tonight if ever it would help.)

gares:
I guess the file should be renamed and the pr merged as soon as Coq/coq#13563 is

PR #35, opened 2018-03-26
RFC: unifying the behavior of reduction tactics and developing a richer language to control the reduction.

herbelin:
Rendered [here](https://github.com/herbelin/ceps/blob/fined-tuned-reduction/text/fine-tuned-reduction.md). I also tried to make a short summary and motivation.

### DIFF
--- a/text/fine-tuned-reduction.md
+++ b/text/fine-tuned-reduction.md
@@ -1,0 +1,45 @@
+- Title: Fine-tuning the language to control reduction / uniformization of reduction behaviors
+- Driver: Hugo
+- Status: Draft
+
+## Summary and motivation
+
+- There are a few discrepancies between reduction strategies
+- While doing proofs, one often needs ability for more fine-tuning of the reduction
+
+This proposal is about erasing some discrepancies and about developing a richer language of commands to control reduction.
+
+## Examples of discrepancies in reduction
+
+```
+Eval cbn delta in let x := 0 in x. (* x unfolded *)
+Eval cbv delta in let x := 0 in x. (* x not unfolded *)
+Eval lazy delta in let x := 0 in x. (* x not unfolded *)
+```
+
+## Proposal
+
+### Extensions of the language for fine-tuning reduction
+
+- To add a variant of ζ restricted to a list of name, so that e.g. `eval zeta [a] in let b:=1 in let a:=0 in a+b` contracts only `b`
+
+- To add a variant of δ for local definitions restricted to a list of name, so that e.g. `eval delta [b] in let b:=1 in let a:=0 in a+b` unfolds only `b` (and keeps the `let`)
+
+- To have this variant support occurrences, e.g. `eval delta [a@1] in let b:=1 in (let a:=0 in a+b, let a:=1 in a+b)` unfolds only the first `a` (and keeps the let-in)
+
+- To delimit a subexpression to evaluate by a context, as in `eval lazy subterm X of pattern [?X + ?] in ((2 - 1) + 3) * 4`
+
+- To restrict a reduction to weak-head reduction, as in `eval whd lazy subterm X of pattern [?X + ?] in ((2 + (x + (2 * 2))) + 3) * 4`
+
+- To restrict a reduction to a number of steps of recursion as in `eval whd lazy fix [3] subterm X of pattern [?X + ?] in ((4 + (x + (2 * 2))) + 3) * 4` which would give `((1 + S (S (S (x + (2 * 2))))) + 3) * 4`
+
+Note that the concrete syntax is pretty speculative yet.
+
+### Proposal for δ on local definitions
+
+To always unfold local definitions whenever δ is expected.
+
+## Related issues
+
+To be eventually able to unfold `fix` and `cofix` into their named variant when such a name exists.
+


maximedenes:
I think it's a very good idea to start a CEP on this! I can't contribute much at this point, except for the weak head evaluation strategy for which I already had a prototype implementation: https://github.com/maximedenes/coq/tree/weak-cbv

herbelin:
Cool! So, let's at least contribute to the design here.

aspiwack:
A few thoughts to make this more helpful and precise.

- All of the bullet points have an example of syntax, accompanied with a short English description of the effect. You should add the expected result in Coq syntax as well. You've done it for the last bullet point, it would be better if all items were like this. (I would also suggest putting the example in code blocks, rather than inline code. For readability)
- Furthermore I think that most of the examples should be accompanied by what the reduction would be like without this feature (for instance the `whd lazy` should be accompanied by the same example, but with just `lazy`)

gares:
I've a question. I see `reduction delta [a] in let a := t in t`.  AFAIK the let-in on the right is in De Bruijn notation, while "a" on the left is a name. How can you implement your proposal?

gares:
Yes but it interacts badly with Ltac.  If you do `Ltac foo t := reduction delta [a] in t` what is `a`?

Either the names are already in your proof context, so that you can pass both `t` and `a` to `foo`,
or the names are not there, hence you cannot use them. The name that is store in the `Lam` node is a pretty printing hint only, it has not to have any meaning (observable by CIC or Ltac).

ppedrot:
> The name that is stored in the Lam node is a pretty printing hint only, it has not to have any meaning (observable by CIC or Ltac).

If only! There are already many places in Ltac that observe this name, amongst others `destruct`, `intros` and the like. For instance,
```
Goal forall n : nat, True.
Proof.
induction n.
Abort.
```
doesn't work, but the following does:
```
Goal forall n : nat, n = n.
Proof.
induction n.
Abort.
```
That we allowed that baffles me.

gares:
Well, I guess we agree then...

gares:
This is even more fun: the name *is* there, but the system uses what the pretty printer would print... amazing!!!

ppedrot:
> This is even more fun: the name is there, but the system uses what the pretty printer would print... amazing!!!

Yeah, this is actually even a source of slowness in a few rare cases.

aspiwack:
Another discussion of interest that this proposal raises: there is a language of subterm selection and such in `rewrite strategies`. How does it compare with this proposal?

gares:
I guess the question is for @mattam82

herbelin:
> Yes but it interacts badly with Ltac. If you do `Ltac foo t := reduction delta [a] in t` what is `a`?

If you write this, `a` is interpreted the same way as an evaluable term is currently interpreted in Ltac. If it exists in the context as a constant (or section variable), it is disambiguated as such. I don't want it to be interpreted dynamically as a possible name in `t`, i.e. it has to be bound statically.

Let's consider alternatively `Ltac foo t := let a := fresh in let t := constr:(let a:=1 in a) in reduction delta [a] in t`. A possible semantics would be to accept this because we can statically check that `t` has a name referred to by `a` in it (whatever exact internal name has been chosen for `a`).

In any case, is the problem really different when referring to an occurrence by name than when referring to an occurrence by another way than by name. If I refer to an occurrence by position, say as in `Ltac foo t := let a := fresh in let t := constr:(let a:=1 in a) in reduction delta [LET at n] in t` (for `LET` a conventional name to refer to a `let`), what do I do? Do I require a static check that this occurrence makes sense (we don't do it as of today for `at n` in `unfold c at n`)? Do I accept to fail dynamically if it does not exist? Do I forbid such construction in Ltac?

> Another discussion of interest that this proposal raises: there is a language of subterm selection and such in rewrite strategies. How does it compare with this proposal?

I don't know well this language and I was not yet at this stage of refinement when I wrote the cep (maybe RFC was not the right header actually).  The language designed by ssreflect developers for denoting (non-binding) occurrences is also a direction on which to capitalize.

> This is even more fun: the name is there, but the system uses what the pretty printer would print... amazing!!!

@gares, apparently, you don't like this convention. And I guess that you don't like either that the induction name is `n0` here:
```coq
Section A. Variable n : nat. Goal forall n, n = 0.
```

I think we should not mix different problematics. From an interactive point of view, I don't find it absurd that what matters is what we see more than what we wrote. This starts to be problematic when we want to have a static view at documents, which is also perfectly legitimate. This starts to be problematic when we want to program tactics, because we want to enforce static properties for them.

Let's list other issues:
-  how to refer to the binding names here:
   ```coq
   Definition P p := forall n, p + n = 0.
   Goal forall n, P n. unfold P.
   (* forall n n0 : nat, n + n0 = 0 *)
   ```
- or here
   ```coq
   Notation P p := (let n := 1 in p + n = 0) (only parsing).
   Goal forall n, P n.
   (* forall n : nat, let n0 := 1 in n + n0 = 0 *)
  ```

ejgallego:
> And I guess that you don't like either that the induction name is n0 here:

That's certainly a bit counter intuitive for folks used to functional programming. In typical Haskell code for example the outer `n` would be shadowed, [and thus inaccessible] and the printer may indeed be ambiguous but that's the programmer's fault, at least it tends to parse back.

aspiwack:
> The conclusion is, imo, that we should not use names to drive reduction under binders. So let's find another way...

This is a bit strong a statement, I think. But anyway, this is quite orthogonal to the proposal under discussion. If someone has a good idea how to better address bound variables in Ltac (and friends): let them make a CEP. It will be most welcome!

It seems safe to me to say that this proposal doesn't do anything new on this particular front (it just expands an existing scheme, and if another one were to exist, it would readily be compatible).

Let's focus this conversation on making this proposal more precise. Let's try and find what parts of the CEP are still fuzzy, unclear, underspecified, what bad interaction there could be with the rest of the system, consider alternative designs and try to catalogue arguments for and against. This is, in my experience, the best way to use this kind of forums.

gares:
> This is a bit strong a statement

Take the lambda term (\x.x + 3). Using x *outside* the binder scope makes no sense. FTR, the binder scope is what follows the . up to the ) . Especially if "x" is not stable (e.g. it could become x0 in a context were x is already used).

> It seems safe to me to say that this proposal doesn't do anything new on this particular front 

*No*.  The argument "we are already doing it wrong" is just wrong.

Going back to the CEP, an horrible, but sensible, replacement is a path from the root to the term you want to point at.
An example is "cross a lambda, first argument of an application". An AST could be "bind(app(1))".
This is the standard way of pointing to a term in all ATP I've seen. I think rewrite strategies do have something like that.

ejgallego:
> an horrible, but sensible, replacement is a path from the root to the term you want to point at.

Randy Pollack had at some point a representation that was exactly that [IIRC he generated it from a duality from the usual deBruijn one for binders]. It was a bit tricky but no so "horrible".

maximedenes:
> it just expands an existing scheme

Sure, Ltac already does it, so let's do it more. What could go wrong?

What could go wrong?

aspiwack:
> No. The argument "we are already doing it wrong" is just wrong.

I think you will all easily agree that this was not my argument. My point was that, unless there is a specific reason for this particular feature to do things differently than the rest of Ltac, this is not the place to discuss such a change.

It is clear that you don't like this mode of addressing. This is a good thing! And you should make a CEP with a proposal for addressing throughout Ltac. This would be really great! If there is not enough ideas to make a CEP yet, you can also open an issue to track your thoughts. If anything this would help gather the reason why the addressing is as it is today. Which would be good documentation. And if we find a better proposal: wonderful. Coq will be the better, and we will have actual documented argument why.

The goal of CEP discussion is to foster new ideas in Coq, and help them get in a good shape prior to integration. Not to stonewall proposals. We should all remember that all too often the perfect is the enemy of the good. Let me plead to everyone to stay focused: these Github conversations are not particularly good at discussing several subjects at once.

gares:
I agree, sorry if I misunderstood.

The title of the cep talks about a language to drive reduction, and the example uses the name of a bound variable to do so. Imo the issue I raise is not off topic. 3 out of 6 examples use names of local variables, and the feature being proposed is about controlling the unfolding of these variables.

mattam82:
Rewrite strategies could indeed be used to do selective reduction (you have patterns there and an eval <redexp> strategy) but are not part of cbv/cbn reductions. Actually, could you explain a bit the discrepancies at the beginning of the CEP @herbelin ?

Zimmi48:
@anton-trunov and @RalfJung: you might be interested in this.

RalfJung:
All of this occurrence and pattern matching very much reminds me of ssreflect rewrite patterns. Ideally, Coq would only have one kind of pattern language used both for rewrites, reductions and whatnot instead of coming up with a new ad-hoc language every time one is needed.

herbelin:
> All of this occurrence and pattern matching very much reminds me of ssreflect rewrite patterns. Ideally, Coq would only have one kind of pattern language used both for rewrites, reductions and whatnot instead of coming up with a new ad-hoc language every time one is needed.

I can only add +1 to the idea of more code, concepts and syntax sharing. Note that, currently, there are at least 3 differents algorithms to select subterms, the one coming with `rewrite_strat` (`rewrite.ml`), the one used for `simpl`, `destruct`, ... in `find_subterm.ml`, the one from the `ssreflect` plugin.

I don't know if @mattam82 has some plans to coordinate a reflection on how to unify and unite these different approaches and get the best of them, and if not, if someone else would be willing to do so (@gares, or @maximedenes, maybe?).

PR #42, opened 2020-02-20
New CEP on unifying the syntax of Theorem, Definition, Fixpoint, etc.

Zimmi48:
[Rendered](https://github.com/Zimmi48/ceps/blob/unify-theorem-definition-fixpoint/text/unify-theorem-definition-fixpoint-syntax.md)


SkySkimmer:
I think I would prefer Theorem (and Lemma etc) to be really an alias of Definition, ie when body is provided you get a transparent constant by default and have to use the attribute to get opacity for both Theorem and Definition.

Zimmi48:
Thanks for this perspective! I hadn't considered it. I'll add it to the "alternatives" section for now.

gares:
I like this. 

Wrt the opaque attribute, you could have a sub attrite, eg opaque(tactics) or better unfold(37) or unfold(never), that is another attribute for the tactic level thing, as one could have simpl(never). In some sense, it may the good occasion to disambiguate, and use opaque only for the kernel notion of opacity

Zimmi48:
And so aim to deprecate the `Opaque` / `Transparent` commands? What about the `Hint Opaque` / `Hint Transparent` ones?

SkySkimmer:
>And so aim to deprecate the Opaque / Transparent commands?

No, changing back and forth is still useful.

text/unify-theorem-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,193 @@
...
+
+We could also decide to implement this plan but make `Definition` and
+`Theorem` fully interchangeable.  A `Theorem` with a body would be
+transparent by default, an `opaque` attribute could be used on
+`Definition` or `Theorem` to make the body opaque.
+
+# Unresolved questions
+
+The name of attributes `opaque` and `transparent` may be confusing
+with the `Opaque` and `Transparent` commands which are not as strong.
```
JasonGross: Let me repeat again my call from six years ago to have more notions of opacity: https://github.com/coq/coq/issues/3389
This is probably beyond the scope of this PR

text/unify-theorem-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,193 @@
...
+    `with` variant of `Definition`;
+  - using the `with` variant of `Theorem` without attribute
+    `recursive` or `corecursive` raises a deprecation message and is
+    interpreted as if `recursive` had been used;
+  - the `@fixannot` clause is reserved to the `recursive` attribute
+    being used (or implicit in the case of `Fixpoint`).
+
+- `opaque` and `transparent` to declare an object opaque or
+  transparent like `Qed` and `Defined` do (note that this is not the
+  same as what the `Opaque` and `Transparent` commands do):
```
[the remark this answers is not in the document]
Zimmi48: This is the question I'm asking in the "Unresolved questions" section.

text/unify-theorem-definition-fixpoint-syntax.md
Zimmi48: That's interesting and also echoes the comment of Enrico above. But yes, this is out of scope.

Zimmi48:
It seems like the main questions to resolve will be about the opacity attribute(s).

Considering the suggestions by Enrico and Jason to extend the available opacity options, it would be great to ensure that whatever is decided is compatible with these views for the future of opacity.

Regarding the question of the default opacity (no attribute), is there any opposition to Gaëtan's suggestion for the default to always be transparent when a body is provided (even with the `Theorem` family)? I am starting to lean in this direction as well since it makes things even simpler to explain and remember.

gares:
>  default to always be transparen

Are you saying that `Theorem x : T. exact t. Qed` is opaque but `Theorem x : T := t.` is transparent? Please no.

Theorems are opaque, Definitions are transparent. How can it be simpler than that?
And I'd take the occasion to disallow `Defined` to end a `Theorem` and `Qed` to end a `Definition`, making the opacity be statically and consistently be known the beginning of the command (that is also where the attribute to alter the default would be).

SkySkimmer:
This is dependent types, proof relevant theorems are normal.

gares:
How does you proposal help wrt this?

text/unify-theorem-definition-fixpoint-syntax.md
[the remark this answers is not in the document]
Blaisorblade: IMVHO, I think `qed` and `defined` would be strictly better names in the current scenario.
Otherwise, please pick *fresh* words that can be used consistently, and make sense if coq/coq#3389 is accepted.

Ideally, technical terms and their meanings would be in a one-to-one mapping, absent strong reasons for a specific case.

Zimmi48:
> And I'd take the occasion to disallow `Defined` to end a `Theorem` and `Qed` to end a `Definition`, making the opacity statically and consistently known at the beginning of the command (that is also where the attribute to alter the default would be).

IIUC, in your proposal, the `transparent` or `defined` attribute would be required to end a `Theorem` with `Defined` (and conversely the `opaque` or `qed` attribute would be required to end a `Definition` with `Qed`). That would be a consistent choice with making the `Theorem` command with a body opaque by default. It would also mean not having to scroll to the end of the proof to check the opacity, which could be an advantage. However, I bet that this wouldn't please the HoTT people. A lesser version of this would be to create a warning (of the `pedantic` category) in case one ends a `Theorem`—without the `opaque` / `qed` attribute—with `Defined`. Then HoTT users could choose to disable this warning. This would be more consistent with the current flexibility of the language.

gares:
No, I meant that `Theorem` ends with `Qed` and `Definition` with `Defined`. By default the first is opaque, the latter transparent. If you give `#[opaque]` to `Definition` you still have to end it with `Defined` (the attribute is supported by the command that begins the thing, not the terminator).

gares:
Wrt HoTT-like users, I don't see how this CEP or the proposal of Gaetan helps.
They want things to be transparent, not a different syntax to make things transparent.
Am I missing something?

tchajed:
I like this, too! In particular it adds a new feature of opaque theorems with explicit bodies using tactics-in-terms (last I checked `Proof with ltac:(foo).` doesn't work, and it's clunky regardless). This allows really concise theorems when they're generated systematically by automation, particularly when the statement itself can be generated by the automation.

I don't have a strong opinion on what to do about the various combinations, but I would like at least a recommendation on what sane developments should do to reduce the many options Coq gives. I sort of agree with @gares that `Theorem foo := t` being transparent is confusing. To me it's the difference between the explanation "theorems are opaque and definitions are transparent", as opposed to "explicit bodies are transparent and proof mode constructions are opaque unless terminated with `Defined`" (but with the caveat that Theorem and Definition are now truly synonymous).

JasonGross:
Perhaps there should be options `Set Default Theorem With Body Opaque` (perhaps with a better name), and a similar option for `Definition`. And then we can document then uniformly (each kind of definition either looks to the theorem or to the definition setting), and differences are only a matter of default settings?

JasonGross:
> No, I meant that Theorem ends with Qed and Definition with Defined. By default the first is opaque, the latter transparent. If you give #[opaque] to Definition you still have to end it with Defined (the attribute is supported by the command that begins the thing, not the terminator).

@gares what should end `Fixpoint`?  And what about `Let`?

SkySkimmer:
>proof mode constructions are opaque unless terminated with Defined

There's no default unlike what your phrasing implies.

I would just say "Theorem is the same as Definition, you can make the `:=` form opaque with an attribute, proof mode opacity is according to how you end it"

gares:
I'd say Defined since it is the most common case AFAIK.

Zimmi48:
> No, I meant that `Theorem` ends with `Qed` and `Definition` with `Defined`.

If we imposed this new view, this would be a pretty big change: everything in HoTT is currently `Theorem` + `Defined`. It would need to be changed to `#[ transparent ] Theorem` + `Qed`. That would also be confusing to everyone who was used to the meaning of `Qed`. While this could make sense for a new language, I feel like this is too radical for an evolution of an existing one. OTOH the proposal in https://github.com/coq/ceps/pull/42#issuecomment-589765428 still makes things closer to your idealized view, while being less disruptive.

Zimmi48:
More precisely @gares, here is a possible path to arrive to what you have in mind:

1. Provide new `transparent` / `opaque` attributes and `Set Transparent Theorems` flag. Deprecate the use of `Defined` for a theorem without the attribute nor the flag.

2. Turn the deprecation into a hard error. Now all Coq projects have either set the flag on, or are using the `transparent` attribute whenever they use `Defined`.

3. Make `Qed` and `Defined` synonyms. Opacity is controlled only by the attribute or the flag. Encourage the use of `Qed` for theorems and `Defined` for definitions. We never have to go beyond this step. Keeping the two things synonyms is in line with the philosophy of Coq (similar to having `Parameter` and `Axiom` synonyms but encouraging their usage in different cases).

Leaving several versions (but at the very least one) between each step should give people plenty of time to adapt. Providing a flag would ensure that HoTT people do not have a hard time because of this change.

gares:
I like your plan. Somehow I recall I've seen hott files with Definition/Defined. I ve just checked a few and they look like that. We're did you see Theorem/Defined used extensively?

Zimmi48:
> Where did you see Theorem/Defined used extensively?

At least in UniMath there is a mix of `Theorem` + `Qed` and `Theorem` + `Defined`. I just assumed this was the same thing for HoTT.

Zimmi48:
I've added the topic to today's call.

Regarding transparency attributes, one option that would be compatible with separating the transparency question from the `Qed` / `Defined` keywords and a larger spectrum of transparency levels would be:

- `sealed` for the current meaning of `Qed`
- `transparent` for the current meaning of `Defined` with the possibility of adding later sub-attributes to specify the precise level of transparency.

Zimmi48:
The conclusion of the Coq Call is to go with the plan at https://github.com/coq/ceps/pull/42#issuecomment-590008625 but with attributes `sealed` / `unsealed` to avoid any kind of confusion with the current `Opaque` / `Transparent` commands and the planned meaning of `Qed` / `Defined` (just a way to close a theorem / a definition regardless of its transparency). I'll integrate this plan in the CEP and then it will have to be largely advertised (Coq-Club, Discourse, Twitter?) before being definitely approved.

ejgallego:
Sorry for missing the call today; it seems to me that removing the current mechanism will have a huge impact in terms of compat; on the other hand I don't see any technical advantage on knowing that a proof is opaque / transparent at proof opening time.

Thus IMO we may weight the cost/benefit ratio here.

Zimmi48:
My own estimation of the cost of porting is "pretty small". Indeed, there will be the option of using the attribute selectively or the flag if all theorems in a file / project must be transparent. As for replacing `Defined` with `Qed`, this will never be forced upon the users since in the end they will become synonyms.

This is why I suggested this plan following Enrico's remarks. I wouldn't have proposed any plan that was strongly disruptive for little gain.

ejgallego:
> My own estimation of the cost of porting is "pretty small".

Well, just the amount of documentation / tutorials, stuff like SF that needs to be adapted seems to be more than "pretty small" to me; but my question really is "why"?

Changing the semantics of current code does not help with any problem I can see, on the contrary, that seems to me like a hack in order to workaround bugs in upper layers. As with the other interpretation restrictions I must stay on the technical view that lower layers should not add workarounds for problems in their clients.

gares:
Note that the design does not require to change existing scripts. It happens that having consistent syntax and attributes one can make documents more coherent in their syntax. Should it become an error, a warning, or nothing, to have Definition end with Qed, is orthogonal to the core of this CEP.

I agree knowing early if a proof is opaque or not brings no advantage to Coq, as well as supporting := in Theorem is not going to change a users life. Having a uniform syntax and a uniform way to control opacity is just making things easier to explain.

SkySkimmer:
>Note that the design does not require to change existing scripts.

I don't see how that's possible, if Qed/Defined don't determine opacity you have to replace them somehow no?

text/unify-theorem-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,193 @@
...
+            | Example
+            | Fixpoint
+            | CoFixpoint
+            | Theorem
+            | Lemma
+            | Fact
+            | Remark
+            | Corollary
+            | Proposition
+            | Property
```
Blaisorblade: What about `Instance`? It seems `Instance` should be a synonym *except* for instance registration. Instead, `Instance` supports a random different subset of features. In particular, mutually recursive instances currently give a parse error (https://github.com/coq/coq/issues/7913).
Hopefully the difficulty of implementation is a separate concern.

ejgallego:
One way we could indeed improve the vernac with less disruption would be to version it; in that case I'd be OK with having better bracketing of vernacs and a few other things, as long as the user could just add a pragma "vernacular syntax 8.13"

maximedenes:
> One way we could indeed improve the vernac with less disruption would be to version it; in that case I'd be OK with having better bracketing of vernacs and a few other things, as long as the user could just add a pragma "vernacular syntax 8.13"

Yes, for vernac (unlike lower layers), it seems feasible.

ejgallego:
> Yes, for vernac (unlike lower layers), it seems feasible.

Other layers do seem feasible too, for example tactics or unification. Pretyping / engine is way more complex indeed.

gares:
I mean: If you have a nicer way to say that a theorem is transparent you can tell the user hey, this looks nicer, please use it. In this case the feature is not going away, so this is what deprecation is about. We can pick another term term if you think it is misleading. Also, if you think we should give n years and not six months to them to get used to attributes, I'm more than fine. Even if n is infinity I m ok.


Anyway, this is not what the CEP is about.
I explained how I see things because your proposal of making Theorem := transparent is very misleading, and clearly conflicts with that vision. If one could not write := after Definition, and you were introducing it in s pr, would you make it produce an opaque term by default?

SkySkimmer:
>Even if n is infinity I m ok.

I definitely don't want to keep something deprecated forever.

>If one could not write := after Definition, and you were introducing it in s pr, would you make it produce an opaque term by default?

I don't think so.

text/unify-theorem-definition-fixpoint-syntax.md
Zimmi48: Indeed, `Instance` should be merged as well and be the same as `Definition` with an instance registration attribute.

Zimmi48:
>Well, just the amount of documentation / tutorials, stuff like SF that needs to be adapted seems to be more than "pretty small" to me

This claim seems to show that you didn't understand what would need to change. For most users, the change would be zero. Please properly read the proposal before commenting it and properly explain what issues you notice with it.

> a hack in order to workaround bugs in upper layers

I don't understand which bugs you are talking about. Please be more specific.

> One way we could indeed improve the vernac with less disruption would be to version it; in that case I'd be OK with having better bracketing of vernacs and a few other things, as long as the user could just add a pragma "vernacular syntax 8.13"

Seriously folks, we are not talking about large disruption here. We are talking about something of a much smaller scale. If you cannot acknowledge this, then I'm backtracking on this part of the plan (which is not yet part of the CEP). The rest of the CEP seemed to be pretty consensual until very recently.

SkySkimmer:
>This claim seems to show that you didn't understand what would need to change.

Could you update the CEP with the plan? That may be easier to understand than a series of comments building on each other.

Zimmi48:
Yes I'll do this tomorrow.

text/unify-theorem-definition-fixpoint-syntax.md
JasonGross: Note that unlike every other definition-like command, `Instance` supports `:= { ... }` syntax for using bidi typechecking for record bodies/types.  (I hear this syntax is going to be deprecated someday?)

Zimmi48: I was aware of this syntax but not of what it did. @maximedenes recently introduced some generic bidi typechecking features (in the `Arguments` command). Are they enough to replace this syntax? Should we also provide an attribute for this?

JasonGross: I am not sure if `Arguments` is enough.  I think it might be almost enough, if you have an attribute that automatically sets it locally at the point in the record constructor between the last parameter and the first field.  However, I think there's also some magic where you can use the unqualified version of the field names even if the module the record is in isn't imported, or something like that (I think similar to how you don't always need to qualify constructor names in OCaml matches).  This bit of it seems more fine to deprecate, though, if need be.

Note that this only makes sense for definitions / instances constructing records.

JasonGross: Also, another difference: `Instance` does not require a name (much like `Goal`), while all the other definition-like commands do.  That is, `Existing Class True.  Instance : True := I.` is valid, and will generate an automatic name for the instance.

JasonGross: > Should we also provide an attribute for this?

I would not object, and it seems like it might be useful.  Something even more useful would be `Set Automatic Bidi Constructors` or similar that made it so that all inductives declared while this flag is set have automatic bidi declarations for all of their constructors, which place the bidi location between the last (uniform?) inductive parameter and the first constructor argument.  I'll go report a feature request for this.

JasonGross: https://github.com/coq/coq/issues/11702

text/unify-theorem-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,193 @@
...
+  transparent like `Qed` and `Defined` do (note that this is not the
+  same as what the `Opaque` and `Transparent` commands do):
+
+  - `transparent` is assumed by default for `Definition`, `Fixpoint`,
+    `CoFixpoint` but `opaque` is accepted;
+  - `opaque` is assumed by default for `Theorem` and variants but
+    `transparent` is accepted.
+  - these attributes are only accepted when a body is provided (when
+    no body is provided the transparency is determined by the ending
+    `Defined` or `Qed`).
```
aspiwack: A fun alternative, which I don't think has been considered in the discussion yet, is to force `Defined` or `Qed` to align with the declaration attribute. It does break some code, though. But it could be considered. Then, to have a transparent theorem, you would need to do:

```coq
#[transparent] Theorem .
Proof.
…
Defined.
```

This may be useful for asynchronous evaluation. Or maybe not, I dont know.

text/unify-theorem-definition-fixpoint-syntax.md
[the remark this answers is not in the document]
gares: FTR at the coq call I heard `seal` and `unseal`, for flipping the "Qed opaque" status. We could then keen `opacity(number)` for the "other opacity".

text/unify-theorem-definition-fixpoint-syntax.md
gares: >  to force `Defined` or `Qed` to align with the declaration attribute

It was proposed to force Theorem to end in Qed and Definition with Defined, have the former default to opaque, the latter to transparent, and have the attributes to overrule the default. @Zimmi48 seem to like it and has a plan to slowly introduce it. It was discussed at the Coq call, sorry you could not attend.

aspiwack: > sorry you could not attend

I don't usually :slightly_smiling_face: I just lurk in random CEPs.

Zimmi48: @aspiwack: I'll push an update today that will summarize the result of the discussion in the thread and in the call.

Zimmi48:
I've pushed the commit integrating the changes to the CEP following discussions in the thread and in the call (new optional part on the interaction of the `sealed` / `unsealed` attributes with the interactive proof mode).

text/042-unify-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,305 @@
...
+
+- The current `Theorem ... with` syntax has been criticized for
+  allowing proofs that are rejected at `Qed` time (unless the
+  `Guarded` command is used).  The problem is exactly the same as when
+  using the `fix` or `cofix` tactics.  When using `Fixpoint` in proof
+  mode, it would seem that the `@fixannot` clause is ignored.  We
+  could look into making this clause being used and the use of tactic
+  becoming safer, but IMHO as a first step it suffices to properly
+  document this behavior and possibly warn when using the `@fixannot`
+  clause that it is ignored.
```
aspiwack: As far as I know, the clause isn't ignored: it's only applied during `Qed`.

It's not easy to check the guard condition on partial terms: they're partial term, they are not really guarded yet.  You could have a partial rejection algorithm. But you have to be careful for it to be somewhat incremental, rather than checking the entire term every time a refinement occurs.

At any rate, this doesn't look particularly related to the current proposal. I'd just remove this paragraph.

Zimmi48: I put it there because @maximedenes was wary of extending further a syntax that he doesn't like.

maximedenes: The problem of this syntax is not described accurately. What I complained about is the syntax *without* the clause.

Essentially, it allows to say "I'll prove this by induction" without saying induction on what (which is not exactly good mathematical practice btw). But this is not supported at the kernel level, so the system generates an ill-typed term. You can observe it by the fact that `Guarded` will fail at any step in the proof. Then, once you reach `Qed`, another hack kicks in to massage the term and finally guess what the induction was about, before sending the final term to the kernel.

Bottom line: this syntax (`Lemma ... with` without a `@fixannot` clause) reflects a mathematically dubious notion ("induction on something"), that is not supported by the kernel. The symptoms are dual to the ones you mention (`Guarded` fails everywhere, proof terms are ill-typed, but `Qed` succeeds magically).

Blaisorblade: `@fixannot` is `{ struct ident }`, right? That's very much not ignored, and it can greatly decrease `Qed` time on big theorems (here, from 8 to 3 seconds) — so much so, my fundamental theorem is a `Fixpoint... {struct }.` (Writing a custom induction principle can be much more expensive).

It could be good to have a warning for the unsuspecting users running into this for the first time, as long as one can disable it :-)

Blaisorblade: > You can observe it by the fact that `Guarded` will fail at any step in the proof.

IME, no; it only fails once you use an assumption on something that isn't a subterm. When I write those proofs, I regularly use `Guarded` (just checked on 8.10, also experienced on 8.9 and probably 8.11). My experience was FWIW on mutual fixpoints.

> Essentially, it allows to say "I'll prove this by induction" without saying induction on what (which is not exactly good mathematical practice btw).

That hasn't stopped Agda from having only this syntax (tho they don't do inlining, so it's cheaper for them). In theory, there is research on cyclic proofs dealing with this.
But in practice, this is just a way to do inductive proofs even when Coq doesn't generate the right induction principles (and for whatever reason you don't want to write your own, or it outright doesn't make sense).

Blaisorblade: > let's spread this syntax to more constructions only once 

"Spread this syntax" doesn't seem a good summary. Right now, all constructs that allow recursion also allow omitting fixannot, regardless of whether they require the proof mode or not — in fact, `Theorem` and variants *forbid it*, which this CEP fixes.

The CEP does add support for recursion to Definition and Example — and allows omitting fixannot there too; so you're arguing to require fixannot there, keeping things irregular.

> we've figured out the mathematical meaning and proper kernel support for it.

I must miss something here too? I'm confident the mathematical meaning is just to infer the annotation.

Your example is clearly a valid bug report (where's the ticket?), and it sounds like the in-progress term should use `{struct ?E}` instead of `{struct y}` where `?E` is a brand-new evar. I feel that should not reach the kernel, but that's because IMHO "kernel" should be reserved for Coq's trusted core, where evars are not welcome (but I guess that's not how things are implemented).

But this CEP is orthogonal. Its point (which I agree with) is that the choice of `thm_token` is basically a required comment whose meaning is only informal — but apparently, quite a few people care about those comments (as argued well by Hugo Herbelin and Xavier Leroy in https://github.com/coq/coq/pull/8265).

Currently Coq assumes that the informal meaning is associated with a different subset of features, but this association is designed in pretty inconsistent way, and this CEP recognizes that this association doesn't always work. Plus, I suspect the implementations are inconsistent — I might or might not have observed `thm_token`-depending bugs.

maximedenes: > and allows omitting fixannot there too; so you're arguing to require fixannot there, keeping things irregular.

Yes. I have a problem with arguments of uniformity with broken features. Note that the "feature" I showed above is also currently incompatible with async proofs (it should fail under CoqIDE, although I haven't tested).

> Your example is clearly a valid bug report (where's the ticket?), and it sounds like the in-progress term should use `{struct ?E}` instead of `{struct y}` where `?E` is a brand-new evar. I feel that should not reach the kernel, but that's because IMHO "kernel" should be reserved for Coq's trusted core, where evars are not welcome (but I guess that's not how things are implemented).

Evars designate subterms. Writing `{struct ?E}` makes no sense, at least in the current system. `{struct` expects an integer which designates one of the variables bound by the `fix`. For this notation to makes sense, you'd need to define this new notion of existential variable, and the guard condition when this variable still hasn't been instantiated.

gares:
Coq is very bad in inferring the recursive argument. It blindly tries all of them and the test can be very expensive. Last time I improved the termination check I had to fight slowdowns here and there, and add struct annotations. This is another reason for recommending the struct argument.

This CEP makes it possible to actually write the annotation in some cases, which is an improvement. It also introduce syntax in which the annotation can be omitted.

I guess the discussion is what to do when the annotation is missing, both in the current syntax and in the new one. My position is that Coq should warn. That warning can be silenced by the user by annotating the fix. Coq should suggest, at Qed time, the annotation.

My2c

maximedenes:
The problem of inferring a missing `struct` annotation when a body is provided is different from the one I mentioned, though. It may be hard to get an efficient algorithm, but at least the problem is well defined. What is more problematic is the absence of such annotation for commands that are constructing the body interactively.

Said otherwise, this CEP introduces the `#[recursive] Definition foo x y : T.` syntax.
1. Is this syntax going to work fine with CoqIDE? If yes, how?
2. How will this syntax avoid the problem I wrote above of having `Guarded` fail at each step in some examples that eventually succeed?
.
> This CEP makes it possible to actually write the annotation in some cases, which is an improvement.

I agree, that part is nice. I'm just advocating to not add syntaxes with induction, no `struct` and no body. Or to explain how to support them properly.

gares:
IIRC Ty he problem with async proofs is that the worker et runs a simple QED, while master et uns more code, including the one inferring the struct. Is this the problem?

The implementation of guarded is just silly. IIRC the hack is to use 0 (first argument) instead of None. Guarded cannot know if it should print "I can't say" or error. Am I missing something?

Anyway, IMO the example you provide should emit as warning. Recursive calls for struct. If you are lazy you can still get the suggestion at Qed (in async mode the worker could run the "inference" code).

To sum up, it is good we discuss details in the CEP, this is what they are for. But I also think we should be more than ok to question the implementation of what we currently have, eg Guarded/dummy value for struct, it the async proof business.

I don't see blocking issues, but yes, I see some extra work required in order to make this CEP bullet proof.

maximedenes:
What would be the typing and reduction rules for `fix` with `None` as the struct?

W.r.t. this CEP, there's another related problem IMHO. I think we should provide syntaxes that do safe inductions (for example, it would be nice if `#[recursive] Theorem` was similar to `Theorem ... intros; induction`, a bit like the "automatic introductions" feature. This would provide a higher level interface than the `fix` tactic.

However, IIUC, this CEP plugs a syntax like `#[recursive] ...` into unsafe (unguarded) induction. It seems to me that this choice will just increase future migration costs.

text/042-unify-definition-fixpoint-syntax.md
Blaisorblade: Yes, by "brand-new evar" I meant what you describe.

Blaisorblade:
> W.r.t. this CEP, there's another related problem IMHO. I think we should provide syntaxes that do safe inductions

FWIW, that cannot replace "unguarded" recursion; the current use-cases of recursive theorems are (1) writing induction principles (2) proving theorems that need a one-off induction principle that's not the default one (because, say, your inductive uses a list). In both cases, calling induction automatically would be a regression: the current situation is suboptimal, but induction makes the goal unprovable, which is worse. With the proof mode, I can at least generate the body of my induction principle in a pretty constant amount of Ltac.

maximedenes:
Who said I wanted to replace unguarded recursion? The `fix` tactic is still available.

I understand sometimes you may not have a choice and have to do unsafe things. Btw (2) seems like an XY problem: you'd want Coq to generate good induction principles for nested inductive types.

But how do you go from "you sometimes need unguarded recursion" to "we should map the main vernacular syntax for recursion to unguarded recursion"? IMHO there should be a well delimited set of unsafe tactics, and the vernacular syntax should not call them implicitly (i.e. it should never produce ill-typed proof terms).

Blaisorblade:
Interesting point, and I don't care about the "main" syntax, as long as the replacement is as convenient. Picking appropriate attribute names is important.
Renaming `recursive` to `unsafe_recursive` might be better, but "term fixpoints" have nothing unsafe, and having different attributes would be inconvenient. Maybe the attribute should be called `unguarded_recursion` plus a togglable warning, to free `recursion` for any future plans?

> Btw (2) seems like an XY problem: you'd want Coq to generate good induction principles for nested inductive types.

Yes and no — I don't see a design or even a feature request for better solutions (which seem much harder), and actual custom induction principles are sometimes appropriate. `fix` seems inadequate: How would I get N mutually recursive theorems from the `fix` tactic *without duplication*? Using `Theorem ... with ...`, I only need to prove each case once. I now fear this produces N copies of the proof term, `fix <proofs> for proof_1` ... `fix <proofs> for proof_N`, but at least the tactics runtime isn't duplicated (and maybe the typechecking isn't either?).

The fundamental difference of opinion seems about policy vs mechanism: What you propose is a fine policy, but it shouldn't get in the way when an exception arises. This CEP is good because it *removes* some Coq policies on user code that are plain wrong, such as "theorems don't need `{struct x}` annotations"; any policies should be warnings.

Worse, the current policies are irregular — you need to combine certain orthogonal features? Right now, whether you can is a coin toss (see also: `Program`).

> Who said I wanted to replace unguarded recursion? The `fix` tactic is still available.
> But how do you go from "you sometimes need unguarded recursion" to "we should map the main vernacular syntax for recursion to unguarded recursion"?

Regularity. "Term" fixpoints use unguarded recursion (undisputedly), so language regularity dictates that proof-mode fixpoints do as well, so that I can freely switch between the two, as needed to compensate other irregularities. Maybe I combine proof mode, `refine` and holes for obligations, so I don't care about `Guarded`.

> It seems to me that this choice will just increase future migration costs.

Migration to what?

gares:
>  It seems to me that this choice will just increase future migration costs.

While I can understand why "unguarded fix" is problematic, I don't get how having `#[recursive] Definition` be a synonym of `Fixpoint` can make things worse. Why would people start omitting struct or relying on the "unguarded fix" feature more than before?

What can make things better is a warning, or an error, in case of a missing struct. This can be done even today, and has nothing to do with this CEP.

maximedenes:
It can make things worse by reserving a syntax that could have been mapped to safe recursion. Using something like `#[unsafe_recursive]` would indeed be better (do you want instead `#[recursive]` and `#[safe_recursive]`)?

Note that the whole idea of an a posteriori warning works only for unsafe recursion, because in real recursion, the shape of your hypotheses depend on which recursive argument you pick, I believe.

gares:
Can I summarize your position as follows.

We need two attributes (instead of one) to signal recursion:
- #[recursive] that should error if no struct is provided
- #[unsafe_recursive] should not, maybe warn, if no struct is provided. It should error if struct is given

maximedenes:
Yes, that's a good summary :)

Maybe `#[unsafe_recursive]` can still (optionally) take a struct, and check the final proof term w.r.t. to that struct (which is the semantics today, I believe).

Zimmi48:
IIUC, your proposition to have a safe `recursive` attribute doesn't map to any feature today, so this CEP would only provide the `unsafe_recursive` one. How much chance is there that we do provide this safe `recursive` attribute in the future? If the answer is "not much", then we are trading off consistency for no good benefit.

maximedenes:
> How much chance is there that we do provide this safe `recursive` attribute in the future?

I would say close to 100%.

Zimmi48:
OK, then the only question remaining is how to name the `unsafe_recursive` attribute. My feeling is that this name is not very good because "unsafe" is not the main characteristic here. Probably "unguarded" describes the thing better. Then, we have to consider how to differentiate what this "unsafe" / "unguarded" means as opposed to the feature provided through the [`Guard Checking`](https://coq.inria.fr/refman/proof-engine/vernacular-commands.html#coq:flag.guard-checking) flag, and which could also be accessible via an attribute. What about the following for instance?

- `recursive` (only available in body mode to start with, proof mode support will come later)
- `recursive(unguarded_tactic_mode)` (can fail at `Qed` / `Defined` time)
- `recursive(disable_guard_check)` (no guard check at `Qed` / `Defined` time; unsound).

I'm also afraid of two things with this attribute:
- it becoming too long. While asking users of `Theorem ... with` to add `#[ recursive ]` in front was reasonable, I wonder if people like Xavier Leroy won't complain if it becomes something quite long like `#[ recursive(unguarded_tactic_mode)` ] unless a flag is provided which when turned on allows to reduce this back to `#[ recursive ]`.
- using `unsafe` risks being confusing to users because "unsafe" usually means something stronger, whereas in Coq, this is still safe because there is the checking at `Qed`. Again, I expect that someone like Xavier would be wary of having to write "unsafe" in his code, given that it is then reviewed by people who might be frightened by this.

Note that I haven't @-mentioned Xavier here because the discussion may be daunting (it's already quite long), but we should definitely reach out to him with a summary to get his opinion (maybe after I integrate the result of this discussion in an update of the CEP).

text/042-unify-definition-fixpoint-syntax.md
```diff
@@ -0,0 +1,305 @@
...
+
+- Another command that should be merged with `Definition` is
+  `Instance` (an `instance` attribute could be supported on
+  `Definition` and variants).  However, given that `Instance` itself
+  has several additional peculiarities (supports `{ }` syntax, accepts
+  auto-generating a name for the instance), this could also happen as
+  a later step.
```
Blaisorblade, line 294: As discussed in https://github.com/coq/ceps/pull/42#discussion_r384727546, while fixing that issue, it'd be great to also address https://github.com/coq/coq/issues/7913 by allowing folded fixpoints, including mutually recursive ones — just run into that again.

That comment is now hidden — It'd be good to insert the link in the CEP itself.

Matafou:
Sneaking into this discussion.

I may be wrong but I have the feeling that the attributes before `Theorem` or `Definition` should give information about the "finished" theorem or definition only.

I find it strange that it determines the def/proof mode (guarded or not). I think the choice of a def/proof mode should maybe be determined by something more "inside".
What about a mandatory `Ungarded Proof.` replacing the usual `Proof.` if you want to use things like `fix`?

Blaisorblade:
> I find it strange that it determines the def/proof mode (guarded or not). I think the choice of a def/proof mode should maybe be determined by something more "inside".
> What about a mandatory `Unguarded Proof.` replacing the usual `Proof.` if you want to use things like `fix`?

I suspect that's a valid question for `recursive(disable_guard_check)`...

... but IIUC, `recursive(unguarded_tactic_mode)` already satisfies that criterion; AFAIK, that attribute is isn't about explicit uses of `fix`, but about emulating `Theorem ... with`, which is equivalent to _generating_ a use of `fix`. (Except that, IIUC, `Theorem A1 with A2 ... with An` produces n uses of `fix`, with the same body but selecting different members).

No idea if the `fix` tactic should receive similar attributes, but I don't know if `disable_guard_check` can be made local to a `fix` expression.

Matafou:
> but about emulating `Theorem ... with`, which is equivalent to _generating_ a use of `fix`.

Good point, fix would be used before the `Unguarded Proof` but nevertheless the latter could be made mandatory before using recursive calls (they would be refused otherwise).

```coq
#[recursive] 
Definition foo (x y:nat).
Unguarded Proof. (* mandatory if struct missing and writing recursive calls *)
  case x.
   - ...
   - exact (foo x0 y). 
Defined.
```
or maybe:

```coq
#[recursive]
Definition foo (x y:nat).
#[unguarded]
Proof. (* mandatory if struct missing and writing recursive calls *)
  case x.
   - ...
   - exact (foo x0 y). 
Defined.
```

My point is that the `unguarded_tactic_mode` appearing at definition level seems a bit strange. Users of my definition do care about the `sealed/unsealed` attribute and also about the `recursive` one (because of the different unfolding reduction rules` but they don't care what the definition mode was during the construction of the constant. It is more "inner" information imho, so I would expect to see them somewhere inside the definition bloc instead at its head.

gares:
Hum, in your example the "unguarded" thing happens as a result of running `Definition`, not `Proof`.
So it seems a bit weird to move the unguarded flag there.

gares:
(I'm all for mandatory `Proof`, I'm just saying how things are today)

Matafou:
> Hum, in your example the "unguarded" thing happens as a result of running `Definition`, not `Proof`.
> So it seems a bit weird to move the unguarded flag there.

We can decide that the `#[recursive]` only says that the definition is... recursive, i.e. starts with a `fix`, without switching to some unguarded mode yet. A warning could suggest the `Unguarded Proof.`. Without it, recursive calls would generate an error with the same suggestion.

This may seem a bit redundant but not more than imposing Admitted after an admit :-)

gares:
When you generate the fix constructor you have to put a number, the rec argument. If you don't give the struct annotation we don't know what to put. If we put a dummy then we are already committing to unguarded mode. At least this is my understanding of the issue. Also we don't have a good API to fix that number later on( it is not an evar)

Matafou:
True, but the unguarded keyword would confirm that the user knows the consequence of not giving the rec argument. I really see this as the dual of forcing the `Admitted.` keyword at the end of a proof containing `admit.`: it is a redundant syntactic confirmation of something unsafe.

Blaisorblade:
@Matafou By your argument, the struct annotation also doesn't belong to Definition or Fixpoint, but that ship has sailed.

For the rest, I think the crucial point is that each individual command has user-visible side effects that are observable through interactive processing (both coqtop and IDEs), so that in the current model a command cannot look ahead.

> We can decide that the #[recursive] only says that the definition is... recursive, i.e. starts with a fix, without switching to some unguarded mode yet.

AFAICS, Enrico is suggesting that's not *actually* implementable; it's `#[recursive] Definition foo` that generates the fix command and adds `foo` to the context, at which point it’s callable.

> Good point, fix would be used before the Unguarded Proof but nevertheless the latter could be made mandatory before using recursive calls (they would be refused otherwise).

Since `foo` is in the context, you could maybe make `Proof.` give a warning.
But `Proof` can be omitted! So it seems only `Qed` can give the error you want (a bit like with `Proof using.`, but this time when you don’t use a `Proof.` command). Then, you have things in the context that you can’t use, with all the associated weird behavior, and you learn at `Qed`-time — possibly after wasting hours? I’m sure that’s not what you propose, but I cannot guess what you have in mind. In fact, time to file/find a bug on the existing behavior of `Proof using`.

Blaisorblade:
If we wanted to follow `Set Proof Using Clear Unused.` (which I find better), it seems you could have `#[recursive] Definition foo` add `foo` to the context, and then `Unguarded Proof.` would proceed normally while `Proof` would clear it again and give a warning, and having no `Proof.` command would only give an error at `Qed`. I wouldn’t like that user experience.

@Matafou What do you think? If I’ve constructed a strawman, you’re welcome to correct me.

Matafou:
OK implementing what I have in mind is probably not worth the effort.

Zimmi48:
FTR I like the separation of concerns advocated by @Matafou, but, like the others, I am also concerned that it is not implementable in the current setting. It seems to me that the only proper way of implementing this would be to make the `Proof` command compulsory (as mentioned by @gares). By making `Proof` compulsory, I mean that `Lemma` alone wouldn't open the proof mode but it would wait until `Proof` to do it. How realistic is that model @gares?

If this is a realistic model, then we could consider having only `#[ recursive ] Lemma` (which is new) have this behavior (for the time being).

ejgallego:
In my view making proof mandatory is a step-back, now you need to parse two sentences to get maybe some important information about the proof.

gares:
> How realistic is that model @gares?

This is simple and sane, but would be a bit odd if one starts an interactive definition.

Zimmi48:
> would be a bit odd if one starts an interactive definition

Yes, we'd need an alternative (equivalent) to `Proof`, e.g. `Start`.

>now you need to parse two sentences to get maybe some important information about the proof

By parse, do you actually mean evaluate, i.e. step forward twice instead of once? This is already the model of obligations and I thought people were happy about it.

Note that having such a delay to start proof mode opens up all sorts of possibilities, for instance the ability when proving multiple theorems at once with `#[ recursive ] Theorem ... with ...` to separate the various goals with `Proof of ...`, similarly to what happens with `Obligation`.

ejgallego:
> Note that having such a delay to start proof mode opens up all sorts of possibilities, for instance the ability when proving multiple theorems at once with `#[ recursive ] Theorem ... with ...` to separate the various goals with `Proof of ...`, similarly to what happens with `Obligation`.

That's a good point, somehow I've been having a mental model of equations / obligations more in the "Agda"-style, but I need to think more about that and read this discussion more carefully.

PRs such as https://github.com/coq/coq/pull/12372 put the ML API in a more canonical form; in the sense of the flow of proof information I tend to think that the upper-layers should follow the lower-level dataflow, [otherwise get bad impedance and bugs], so indeed I think this kind of design questions are better addressed first at the ML API level, following the high-level objectives outlines for example here.

Alizter:
I'm not sure if it was mentioned, but it would be nice if we could set what the default opacity for the `Definition`, `Theorem`, etc. keywords were. Currently in the HoTT library, we use `Theorem`, `Definition` etc. interchangeably with transparent bodies. They essentially serve as documentation.

I don't mind the default behaviour of `Theorem` being "sealed", as long as we can change it.

Zimmi48:
Yes @Alizter, this was proposed a few messages above by @CohenCyril.

text/unify-theorem-definition-fixpoint-syntax.md
Blaisorblade: > However, I think there's also some magic where you can use the unqualified version of the field names even if the module the record is in isn't imported, or something like that (I think similar to how you don't always need to qualify constructor names in OCaml matches). This bit of it seems more fine to deprecate, though, if need be.

I’d rather we make it more widely available — field names should always be disambiguated by type even when not imported…

Zimmi48: @Blaisorblade Could you open another issue for this? It should probably be tracked outside this CEP.

Alizter:
@Zimmi48 Shall we add this for discussion in the next coq call? I think it would be good to begin working on this.

Zimmi48:
@Alizter there have been some comments in the discussion here that I've yet to integrate to the CEP. But the most problematic for the advancement of this CEP is that I'm lacking time to implement it.

herbelin:
A summary would be useful in any cases. Others may find time to implement if needed.

herbelin:
What about already implementing sealed and unsealed attributes? That does not look like a big deal and would provide an immediate gain.

(And if we want to go to massive renamings or deprecations, I don't realistically see how to avoid doing it using an automatic tool. As an example, there are already many warnings in the CI which are not addressed and I don't see how they could reasonably be all addressed manually.)

SkySkimmer:
The idea of choosing sealedness based on if Theorem or Definition was used still feels quite bad to me. Maybe I'm too used to the current system but I do have 1 argument: currently to understand if something is sealed I need to remember the meaning of exactly 2 commands (Qed/Defined) but with this change I would need to know the sealedness of all of Theorem Definition Lemma Example etc

ppedrot:
I agree with Gaëtan that we're spreading state around, and it's always annoying to have to reconciliate disagreeing states.

silene:
As far as I am concerned, I do disagree. I think the head keyword should also be used to set the default sealing state, especially since there is no `Qed`/`Defined` tail keyword when the proof term is directly provided. Having `Definition foo := ...` and `Theorem foo := ...` be perfectly synonymous is a pointless feature for users like me.

ppedrot:
But then how do we know about the ocean of weird "natural-language" keywords that we have in the implementation? And it becomes non-uniform w.r.t. other proof-starting commands like Instance.

herbelin:
As a reference, I copy how `About` reports about sealedness/opacity:
```ocaml
let print_opacity env ref =
  match opacity env ref with
    | None -> []
    | Some s ->
       [pr_global ref ++ str " is " ++
        match s with
          | FullyOpaque -> str "opaque"
          | TransparentMaybeOpacified Conv_oracle.Opaque ->
              str "basically transparent but considered opaque for reduction"
          | TransparentMaybeOpacified lev when Conv_oracle.is_transparent lev ->
              str "transparent"
          | TransparentMaybeOpacified (Conv_oracle.Level n) ->
              str "transparent (with expansion weight " ++ int n ++ str ")"
          | TransparentMaybeOpacified Conv_oracle.Expand ->
              str "transparent (with minimal expansion weight)"]
```
This suggests that a less breaking alternative to `sealed`/`unsealed` could be `fully_opaque`/`transparent`, even more that `Opaque`/`Transparent` are attributes only of `unsealed`, so that, morally, there are not 4 but 3 statuses: fully opaque, opaque only for reduction, fully transparent. If we then accept to mix the kernel opacity and the reduction opacity in the attribute, it could also be, say:
-  `fully_opaque`/`reduction_opaque`/`transparent`??
- `opacity=full`/`opacity=reduction`/`opacity=transparent`??
- `opacity=sealed`/`opacity=opaque`/`opacity=transparent`??
- `opacity=sealed)`/`opacity=reduction)`/`opacity=transparent)`??
- `opacity(sealed)`/`opacity(reduction(level="1",expand,...)`/`opacity(transparent)`??
- etc.

Also, for reference, there are options `Set Transparent Obligations` and `Set Extraction AccessOpaque`. which would have to be renamed accordingly.

ppedrot:
@silene to support our point, here is a list of definition tokens from the grammar. What are their expected sealedness? No cheating.
- Axiom
- Axioms
- Conjecture
- Conjectures
- Corollary
- Definition
- Example
- Fact
- Hypotheses
- Hypothesis
- Lemma
- Parameter
- Parameters
- Property
- Proposition
- Remark
- SubClass
- Theorem
- Variable
- Variables

herbelin:
Sorry, there are parallel discussions and I'm replying here on "mixing two views": it is worrying me also but we also need to take care of common terminological usages in place (including in our own practice). We may try to change the `About` message and the options referring to opacity/transparency in the sealed sense but maybe there are less disruptive solutions that would contribute to cause neither a confusion between the concepts nor a confusion with the existing usages.

herbelin:
Re: default sealedness. @SkySkimmer, @ppedrot: is your proposal that declarations using `:=` (rather than some interactive proof) are unsealed by default, whatever keyword is used?

SkySkimmer:
yes

ejgallego:
It seems to me that just adding general support for `:=` for those commands that have bodies is good. I guess things should work like:

### Default modalities:

- `$logical_kind $name := $body.`: by default not sealed, errors if `$logical_kind` doesn't accept a body.
- `$logical_kind $name. Proof. $body_tac. Defined`: Similar as above.
- `$logical_kind $name. Proof. $body_tac. Qed`: `$name` will be sealed, $body is put in opaquebodies and thanks to typing commands cannot read it.

If the command has an attribute, there are 3 cases:
- the command is of the shape `Proof. ... {Qed,Defined}.`: in this case, if the attribute agrees with the terminator, there is a warning (redundant attribute), if the attribute disagrees, error.
- the command is of the shape `:= $body`: if the attributed is sealed, seal the definition (put in opaquebodies), otherwise warn for redundancy.

I am not sure how this would work, IMHO it is hard to predict more without going and implementing a proposal, and playing with it a bit. The good news is that I think `declare.ml` is in pretty good shape and doing something like the above shouldn't be a huge amount of work.

ejgallego:
I share @silene 's pain point about `Theorem $name := $body.` not being sealed by default. As a user, I would like to have this syntax. Now that I remember, it was in fact the main motivation I think, to have a syntax of the `:=` form that is sealed by default.

ejgallego:
I just read the proposal again (congrats @Zimmi48 , great writing!), I am thinking that maybe we are trying to do too much.

IIRC the main motivation was to have `Theorem foo := ...`. So maybe there are a couple of things to remove from the CEP:
- attributes: indeed, I'm not sure why they are useful here. Coq has inherit the non-uniformity of the syntax `Definition / Theorem` since a long time, so attributes would be only useful if we removed that distinction, but then what's the point of having so many synonyms?
- ability to override the defaults: that can be very confusing for others reading code! Do we have a real case for this customization to be needed?

ejgallego:
> I'm not sure if it was mentioned, but it would be nice if we could set what the default opacity for the `Definition`, `Theorem`, etc. keywords were. Currently in the HoTT library, we use `Theorem`, `Definition` etc. interchangeably with transparent bodies. They essentially serve as documentation.

But you cannot use `Theorem foo := ...` yet, so are you forced to emulate it?

> I don't mind the default behaviour of `Theorem` being "sealed", as long as we can change it.

Yeah, that's one of the disadvantages of this approach :(

But at least, the proposed change wouldn't break your current code.

Zimmi48:
Feature-wise, having the `sealed` attribute, combined with the existing `Definition :=` syntax would be enough to achieve the initial proposal from this CEP.

However, another point of this CEP was to create more uniformity in the syntax. It is my opinion that Coq has too much "accidental" variability and we should strive to remove some. That's one of the motivations for the CEP. I'm also of the opinion that (real) synonyms are fine and consistent with the current "culture" of Coq (without the drawbacks of the accidental variability). That being said, the proposal of Pierre-Marie and Gaëtan is fully compatible with this vision, and therefore I'm also OK with their proposal.

I have no opinion about the value of knowing the transparency of a constant without executing the code. The proposal was indeed that Coq would only provide a default transparency for various commands, but that this could be tweaked. It is my impression that the same is currently true with other properties of definitions (if you don't know what options are activated, you won't fully know how the constant is defined)...

herbelin:
PR coq/coq#19029 experiments adding `sealed`/`unsealed` attributes. Fully renouncing to the role of `Qed`/`Defined` for deciding opacity (as was sometimes discussed) would be quite disruptive. A simple possible rule is the following:
- `#[sealed]`/`#[unsealed]` take precedence over the presence of `Qed`/`Defined` or the presence of `:=`;
- in the absence of `#[sealed]`/`#[unsealed]`, opacity is determined by the use `Qed`/`Defined` for interactive constants;
-  in the absence of `#[sealed]`/`#[unsealed]`, opacity is transparent for all non-interactive declarations currently using `:=` (e.g. `Definition`/`Fixpoint`/`CoFixpoint`/`Coercion`/`Instance` ...).

It seems that the only real pending question is the default opacity to give to a future support of `:=` for `Theorem`/`Lemma`/`Proposition`/`Fact`/`Remark`/`Corollary`/`Property`. I'm not worrying about remembering the default status if ever the default is opacity when using `:=` with these 6 keywords, as this would be only and exactly for these 6 predefined "theorem-like" keywords, no more. Adopting that `:=` implies transparency by default (even with those 6 keywords) makes also sense though.

herbelin:
I don't know if it was mentioned in the discussion. To unify the syntax, we would also need a way to tell that a non-mutual `Theorem` is co/recursive without giving a `struct` (which is anyway not available if intended to be a cofixpoint). E.g. with:
```coq
#[recursive] Theorem f binders : type := body.
```
or
```coq
Recursive Theorem f binders : type := body.
```
This would then be applicable also to `Instance` (as in #7913), `Coercion`, `Canonical Structure`, ...

(Or maybe also adding `#[recursive(fix)]`, `#[recursive(cofix)]` to be more specific.)

herbelin:
To satisfy the three needs given in the motivation of the CEP:
- to be able to define an opaque object without giving the type
- to be able to use `Theorem with`  with only one recursive declaration
- to be able to use `Theorem` with `Qed` even for transparent theorems such as `eq_sym` (for pedagogy (**))

I propose to respectively eventually implement in #coq/coq#19029 the following new attributes:
- `#[sealed] Definition c := b`
- `#[recursive] Theorem c : t. ...` (***)
- `#[defined] Theorem c : t. ... Qed.` (*)

The question of what would be the default of opacity of `Theorem c := b`, if provided, did not lead to a consensus (some people primarily see the word "Theorem" and see it as opaque by default while other people primarily see the symbol ":=" and see it as transparent by default). The best solution is thus probably to renounce to provide this syntax.

(*) in the CEP, the keyword `unsealed` was proposed. The idea to use `defined` instead is somehow a new proposition I'm just doing which looks less disruptive compared to the current status (that is: `#[defined]` tells to behave as a definition).

(**) pedagogy is also flexible: if really needed we can also explain why Coq does not follow the standard conventions.

(***) somehow, something is still missing to tell that a theorem which is guarded both inductively and coinductively should be considered as a cofixpoint.

Would someone be unhappy with this proposal?

herbelin:
The part of the previous comment about `sealed` and `defined` is implemented in coq/coq#19029.

PR #43, opened 2020-03-02
New CEP on refactoring the refman.

Zimmi48:
This CEP comes quite late as work in already ongoing with coq/coq#11601 et al., but better late than never.


gares:
Can you say a bit more about `Writing Proofs`? In particular what are your plans about the ssr chapter?

Zimmi48:
Thanks for asking. First, just to be clear, the detailed outline of each chapter is something that will have to be proposed and debated in specific PRs on the Coq repo.

But to still answer your question, what I have in mind as of today is the following. I think we may want to keep the proof handling chapter as a first section (on its own page) that provides a good introduction (it is not too long today, but maybe it could still be improved a bit), then I mostly intend to split out different sections of the Tactics chapter into multiple pages (cf. a few ideas that I proposed in https://github.com/coq/coq/issues/8946) and to kill the "Detailed example of tactics" chapter which makes no sense today.

Regarding SSReflect, I do not intend any change for now (except maybe a pass to check the grammar), so this would mean that it would still be presented on a single page, which would be one of the top-level sections of the "Writing proofs" chapter (much like the proof handling one).

Longer term it would be great to at least take out the non-proof parts of SSReflect:
* move the parts about the Gallina extensions out of the SSReflect plugin into the main code base (and thus similarly move the doc);
* merge the two Search commands (and thus similarly merge the doc).

And possibly to try to provide more symmetry between the way that vanilla and SSReflect tactics are presented, possibly with lots of cross-references between the two, or even by having sections by type of tactics divided in sub-sections presenting the vanilla version and the SSReflect one.

gares:
The I'm ok with plan. My point is that SSR is a language, with a style, idioms, etc, not a list of tactics and tactic combinators that can be presented in any order, or grouped by domain. Hence I'd like its description to stay in a single place. Seach and Gallina extensions can of course be moved elsewhere.

herbelin:
The proposal if ambitious and very relevant! With all my support.

PR #44, opened 2020-04-12
New CEP on rehauling the vo format.

ppedrot:
Corresponding PR: https://github.com/coq/coq/pull/12092.

### DIFF
--- a/text/044-vo-named-format.md
+++ b/text/044-vo-named-format.md
@@ -1,0 +1,163 @@
+- Named-segment format for vo files
+
+- Drivers: Pierre-Marie Pédrot
+
+----
+
+# Summary
+
+We propose a change in the Coq object file format, which would allow to
+access subcomponents of the file per name instead of per position as done
+currently.
+
+# Motivation
+
+The current Coq object file format, hereafter abbreviated CFF, is used to store
+compiled libraries, typically as `.vo` files, or partially compiled ones, e.g.
+as `.vio` or `.vos`. The format consists of a header,
+followed by a list of *segments*. Segments contain marshalled OCaml data,
+although this is not a requirement of the format *per se*, together with a hash
+and position information. In particular, the contents of a segment is uniquely
+determined by its position in the list of a segment. As of 8.11, the segments
+are, in the file order:
+- Summary of the file
+- Logical contents of the library
+- Delayed universes (occurs in delayed compilation of partially compiled files)
+- Delayed proofs (only relevant for partially compiled files)
+- Opaque proofs
+
+This representation is very rigid. Most notably, adding optional segments
+requires changing all tools manipulating vo files. For instance, the delayed
+proof segment is non-empty only in `.vio` files, but all callers need to be
+aware of it. Another issue is that it prevents having a variable number of
+segments, e.g. all opaque proofs need to be stored in the same OCaml array and
+cannot be stored in a per-proof basis, since the number of opaque proofs
+depends on each file. In turn, this does not allow clever on-disk storing
+schemes of opaque proofs as all proofs need to be loaded in memory in the array
+when marshalling this segment.
+
+Instead, we propose a CFF where segments can be accessed by name instead of
+position. This would solve both problems at once, by making the format at the
+same time more robust and more extensible.
+
+# Detailed design
+
+## Inspiration
+
+We introduce a new CFF inspired by the ZIP data format, following experiments
+with the `camlzip` library. We decided against ZIP because this format suffers
+from historical 16-bit and 32-bit limitations, e.g. a ZIP file can only contain
+up to 65535 subfiles, and weight up to about 4GB. Working around these
+limitations requires the introduction of the ZIP64 variant which is very
+inelegant in order to preserve backwards compatibility with ZIP to some extent.
+The ZIP format also suffers from a few additional quirks (e.g. ambiguities in
+the serialization scheme) that render it improper for high-confidence formats.
+
+## On compression
+
+The experiments also showed that systematic compression would incur an important
+runtime cost. ZIP supports several compression schemes, of which `camlzip`
+implements two, `DEFLATE` (LZ-style, with various levels of compression) and
+`STORE` (no compression). All levels of `DEFLATE` triggered important slowdown,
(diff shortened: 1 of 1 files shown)


ppedrot:
cc @gares

gares:
It looks good to me. I think you assume only one segment can grow on disk as things roll. We need one for proofs, but there are others that are streamed, like glob and aux. They are small, and could be buffered, so I guess it is not a problem but seems to be a limitation worth documenting (esp in the API to write to the format)

text/044-vo-named-format.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Overview
+
+The major difference introduced by the new COFF is a metadata registry at the
+end of the file that contains information about the segments, and in particular
+their names. Merely parsing this metadata is enough to know the position of
+a segment inside the file. The position of this registry is known by parsing
+the header of the file.
+
+Another design choice is the reliance on 64-bit integers to indicate file
```
ejgallego, line 77: @corwin-of-amber , how would this impact jsCoq ? Also, quite a few number of users use 32bit Coq as it  does provide in general a non-trivial speedup; we should measure the impact here.

We could allow both 31 / 63 bits and add a tag to the file; but indeed I'm split on that.

ejgallego:
It looks good indeed, tho I think it falls short for many uses cases, for example what's the cost of adding new objects?

I tend to prefer formats that are based in binary indexes over hashes and thus act as a pure database, for example I wonder how git's format would work [of dune db is also pretty interesting tho it uses a file-based cache in the end]

Anyways, even if this format may not cover all future use cases it is still internal to Coq so I see no problem updating it.

text/044-vo-named-format.md
corwin-of-amber: Well: if there are arbitrary Int64s in the data structure, and `Marshal` is used to serialize it, then we will have errors when trying to save it with `Compat_32` if the values are too large.

text/044-vo-named-format.md
```diff
@@ -0,0 +1,163 @@
...
+the files. Thus, it should not affect file size much.
+
+# Alternatives
+
+We could have used a standard file format instead of rolling up our own.
+Unfortunately I did not find a format that was satisfying all the following
+constraints.
+- The container could store uncompressed data
+- The data is random-access
+- The format does not suffer easy-to-hit limitations.
```
corwin-of-amber, line 147: What about key/value store implementations such as Berkeley DB?

text/044-vo-named-format.md
ejgallego: If that's the only problem I think we are good for now; I understand values a file offsets so if we are using `.vo` files > 4GiB then we have other things to worry about I think.

text/044-vo-named-format.md
ejgallego: :+1: I think Dune uses a variation of that and it works pretty well indeed

text/044-vo-named-format.md
corwin-of-amber: Yeah, I think we can safely assume that we would not want to distribute such large files over the Web anyway...

ppedrot: I don't know for JSCoq (even though I suspect it's the same), but it shouldn't impact *at all* the standard OCaml runtime in terms of performance. This is merely allocating a handful of words while the deserialization itself can be arbitrarily large. This is really just about being able to seek in big files, it doesn't affect the serialization of the data proper.

ppedrot:
@ejgallego the COFF is orthogonal to content-addressability, if I understand properly what you're asking for. You can store hashes as segment names if you want (and actually I am planning on doing that for VM bytecode). Also, I'd prefer a simple format, we definitely don't need a full-blown database for Coq files...

text/044-vo-named-format.md
ejgallego: @ppedrot javascript doesn't have 64bit integers O_O

ejgallego:
> @ejgallego the COFF is orthogonal to content-addressability, if I understand properly what you're asking for. You can store hashes as segment names if you want (and actually I am planning on doing that for VM bytecode). Also, I'd prefer a simple format, we definitely don't need a full-blown database for Coq files...

Depends on your granularity level, for example if your build system is incremental per-file this kind of format falls short; you really want something based on B-trees at least, in the sense that you'd be talking about 10 of thousands of segments. I plan to have a need for such kind persistence format soon, but I can roll it own my own for now.

The proposal is good now I think, I am merely commenting.

text/044-vo-named-format.md
corwin-of-amber: FWIW, WebAssembly, while being a 32-bit platform, has 64-bit integers. I think that, at least for the sake of interop with WebAssembly, they will be introduced into JavaScript as well, eventually.

ppedrot: @ejgallego How is JScoq interpreting the `int64` type? The C code is not bound to a software-emulated implementation?

corwin-of-amber: Not exactly – js_of_ocaml translates OCaml to plain JavaScript, there is no "virtual machine" other than the JS interpreter. I believe that if the type is defined explicitly as `int64` (rather than `int`), then it *will* be represented reliably using a tuple of smaller integers. I have not tried to use `int64` in this context, though.

silene:
Please do not use COFF as an acronym for this format. For a while, I really thought you were proposing to use COFF as the format for .vo files. And it was making sense to me, since COFF is the format that introduced the notion of named segments for storing compiled objects. (On Linux, ELF has become the standard format nowadays. But on various other operating systems, COFF or one of its variants is still the format used for compiled objects and executable programs.)

ppedrot:
@silene Arf. The name "COFF" is limited to the CEP, it is defined there and does not appear anywhere else. The fact that it clashes with other abbreviations is thus not important.

gares:
I thought  PMP was making a joke :-)

ppedrot:
I've tweaked the acronym, it now reads CFF.

ppedrot:
If there are no further comments on this CEP, I think that the corresponding PR needs an assignee.

PR #47, opened 2020-10-25
Contravariant subtyping and eta-reduction

herbelin:
[Rendered](https://github.com/herbelin/ceps/blob/contravariant-subtyping-and-eta-reduction/text/contravariant-subtyping-and-eta-reduction.md)


gares:
I gave a quick read to the text. To me it looks like a research program (with good chances to succeed, I trust you) rather than a CEP. The point of CEPs is to sorts out the design (here the theory) on paper before proposing the integration.

I'm just curious, are you or your students working on the paper?

herbelin:
> To me it looks like a research program (with good chances to succeed, I trust you) rather than a CEP. The point of CEPs is to sorts out the design (here the theory) on paper before proposing the integration.

Regarding the question of whether I would trust less a version of Coq with contravariant subtyping, the answer is no, as I argued above.

For instance, regarding @barras' formalization of a set-theoretic model of CIC, I don't have in mind the exact missing parts of  the formalization (and I'd be glad to have feedback of someone knowing it well), but, from memory, if there are things to worry, it is typically more about the guard condition than about η-expansion.

I tend to believe that there is some consensus about the consistency of contravariant subtyping. Besises the arguments I gave above, already at TLCA 1997 I had a discussion with Zhaohui Luo about it and he considered that a realizability model (instead of a set-theoretic model) would justify it. This is also the position defended by @barras in the conclusion of his habilitation in 2012 (section 9.3.3).

The view that contravariant (but also covariant) subtyping of Π was reducible to a question of η-expansion was also advocated by Ali Assaf (in a private communication with him, but maybe also in papers).

Note also that CIC has already η in the conversion and that interpreting `f` ≣ `fun x:A => f x` set-theoretically already requires a form of typed conversion (even if η in the conversion is compatible with equivariant subtyping of the domain).

Of course, this view has to be confrounted to the opinion of alternative experts, starting with experts of the MetaCoq formalization of the theory of  Coq: should we have some doubts about the consistency of the change proposed here?

> I'm just curious, are you or your students working on the paper?

This being said, there is indeed a PhD student of @letouzey and I working on making subtyping explicit in the CIC. Eventually, this would be another way to interpret contravariant subtyping set-theoretically, with the marks telling when to coerce the domain of a function in the interpretation.

This has to be discussed of course, but, assuming a consensus on the absence of theoretical risks, I believe that we have more to gain to have a convenient form of η-reduction than to stay with the current situation.

mattam82:
I didn't review the CEP entirely but I have the exact same view in MetaCoq that eta-reduction + contravariant subtyping should be fine syntactically, and that's one of our next steps (a student of @yforster is working on it). I didn't see the CEP before, I was on vacations :)

herbelin:
> eta-reduction + contravariant subtyping should be fine syntactically

Indeed, I don't see anything specially frightening with this CEP. It would instead help to get a smoother metatheory which I see as a good sign.

herbelin:
@ppedrot: can you give a reference to the subtyping indecidability issues you were mentioning at the call. I'd like to understand if it would apply here.

mattam82:
I guess @ppedrot refers to the fact that we need to handle `l <= max(i, j)` constraints potentially. Actually, we could keep the current setting where domains of arities contain only levels (no algebraics, hence no constaint as above) and then I don't see why anything would change performance-wise.

Note that I will rather advocate for a full handling of algebraics in the future. https://github.com/MetaCoq/metacoq/pull/511 provides a proof of the syntactic metatheory of PCUIC extended with them, along with a (naïve) longest-path algorithm for deciding them, which is proven correct. Hopefully the more complex algorithm in Coq can be proven correct as well by adding negative weight edges and little more changes. This means we can really handle `Set <= l <= Set+1` constraints, and arbitrary `l <= max(i, j)` constraints, *during checking*. This is where obviously we introduce a complexity bump due to the disjunctions.

mattam82:
The inference issue is more complicated, because we'll get `?l <= max(?i, ?j)` constraints during inference, for which it is certainly impossible to find a most general solution.

SkySkimmer:
We already handle `l <= max(i,j)` in checking mode though

ppedrot:
Subtyping is a can of worms in general, empirically it's often a source of undecidability for many weak type systems, and contravariance is a key ingredient in this kind of results. As a preliminary reference, I think that https://sci-hub.do/10.1109/LICS.1996.561306 migh be a good start, although I don't know how much of their results apply to CIC + cumulativity.

herbelin:
This Tiuryn-Urzyczyn paper is about subtyping polymorphic types, i.e. ` ∀α.T ≤ T[α:=U]` and `T ≤ ∀α.T[α:=U]` for `α` fresh`. In the case considered above, it is only about simple types, so this seems rather different.

ppedrot:
@herbelin I agree that the setting is quite different, and that cumulativity subtyping considers sorts when this paper considers second-order types. But this CEP is making quite extraordinary claims in terms of foundational properties of CIC, so we should be overly cautious about its consequences. For a starter, I'd like to see a precise way to implement it, and as a silver standard, the corresponding algorithm in MetaCoq + a proof of completeness, but that's probably not something you'll get by tomorrow morning.

mattam82:
@ppedrot don't push me ;) I think the contravariant subtyping part is way easier to formulate than eta-reduction. And indeed, in our case the proposal is "just" about handling `max(i+n, ...) <= max(j'+k,...)` constraints, being contravariant on arrows and using the appropriate variance on cumulative inductive types. I don't expect we'll have first class universe quantification anytime soon, if ever.

@gaetan indeed, the kernel already implements the more general check, but I doubt it is exercised much. That would be good to know though, if we already fall in the `i <= max()` case today. At least the pretyper should error quickly with "algebraic universe on the right".

jeanas:
> As a preliminary reference, I think that https://sci-hub.do/10.1109/LICS.1996.561306 migh be a good start,

(Link is broken, the paper can be found at https://core.ac.uk/download/pdf/82236417.pdf)

PR #52, opened 2020-12-19
[RFC] release cycle

gares:
Here a draft of what I presented at the Coq call, comments welcome.

Rendered [here](https://github.com/gares/ceps/blob/platform-release-cycle/text/052-platform-release-cycle.md)

CC @Zimmi48 @erikmd @msoegtropimc

### DIFF
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -25,12 +25,12 @@ last:
 - in order to test Coq you are likely to need a bunch of Coq packages, and they
   are typically only available after the beta period is over
 
-Moreover the "beta period" is also seen by devs as a time frame where they can
-still change many things. As a results users are even less happy to spend time
-working on a moving target.
+Moreover the "beta period" has also been seen by devs as a time frame where
+they can still change many things. As a results users are even less happy
+to spend time working on a moving target.
 
 Finally, due to the cost of backporting breaking changes from the master
-branch to the release branch, this only happenes very early in the release
+branch to the release branch, this only happens very early in the release
 cycle (for example only 2 such changes were backported in the beta period
 for 8.12, 0 for 8.13).
 
@@ -50,21 +50,22 @@ component of it.
 
 The new process' timeline:
 ```
-                     D             D            D   D  D
+                     DO+           O            O   O  O
 coq      --+---(1)---+------(2)----+-----(3)----+---+--+----
  vX branch/          |             |            |
-            X+rc tag/              |            |
-                           X.0 tag/             |
-                                        X.1 tag/
+           VX+rc tag/              |            |
+                          VX.0 tag/             |
+                                       VX.1 tag/
 
-                                   I                    I
+                                  DI                   DI
 platform  -----------+----(4)------+--------(5)---------+
            vX branch/              |                    |
-                        X+beta tag/                     |
-                                                X.0 tag/
+                      VX.0+beta tag/                    |
+                                              VX.0.0 tag/
 Artifacts:
-- D = docker image for Coq (and opam package)
-- I = binary installers for Coq platform (and opam packages and docker image)
+- D = docker image for Coq (or Coq + the platform)
+- O = OPAM package (O+ means for core-dev, otherwise it is for the main OPAM repo)
+- I = binary installers for Coq platform
 ```
 
 ## Coq
@@ -72,62 +73,82 @@ Artifacts:
 On time based schedule the RM branches vX.
 
 1. The RM shepherds the few PR which are ready and pins projects tracked by CI
-   (using commit hashes, not necessarily tags). Then he tags Coq. This should
-   take 10 days. No packages are built, just a git tag. A docker image is built,
-   so that project maintainers can use it in CI. No breaking change is allowed
-   from now on (unless a severe problem is found).
-2. Doc is updated (eg. Changes file) and eventual fixes required by the platform
-   are done. Ideally no other change is done. This should take 10 days.
+   (using commit hashes, not necessarily tags), then the TM tags VX+rc.
+   This step should take no more than 10 days (2 working week).
+   No binary installers are built and published by the RM, just a git tag.
+   The RM writes an OPAM package, currently the policy is to upload it to
+   `core-dev`.
+   Possibly, the RM builds a docker image, so that project maintainers can use it
+   in CI during (4).
+   **No breaking change is allowed from now on, unless a severe problem is found.**
+2. The documentation is updated (eg. the Changes file) and eventual fixes
+   required by the platform are done. Ideally no other change is done.
+   This step should take no more than 10 days.
 3. In response to a severe bug report Coq devs make an hotfix in master which is
    backported to vX by the RM which then tags a point release, possibly as soon
-   as the fix is available and merged.
+   as the fix is available and merged. The RM writes an OPAM package for
+   the main OPAM repository.
 
 ## Platform
 
-When Coq X+rc is tagged, the PRM branches vX
+When Coq VX+rc is tagged, the PRM branches vX
 
 4. Starting with the pins made by RM on Coq's CI all packages part of the
    platform (or its core) are pinned (in accordance with upstreams, which are
    notified about the ongoing process). Most, if not all, packages are in Coq's
-   CI so there should be no surprises, otherwise issues are reported to Coq
-   devs. When done a X+beta tag is done and packages made available to users.
+   CI or the Platform's CI so there should be no surprises.
+   When done a VX.0+beta tag is done and packages made available to users.
    A docker image with the entire platform prebuilt should be built. This
    should take 20 days.
 5. As users pick up the platform and find severe bugs in Coq, the platform picks
    up point releases of Coq containing hotfixes and eventually extends packages
    beyond the core set.
 
+### Platform versioning
+
+It is out of the scope of this CEP to chose a versioning schema for the
+platform. In the diagram above we append a .digit to the Coq version (one
+of the current proposals).
+
 ## Synchronization points
 
 - The end of (1) starts the release cycle of the platform.
 - The end of (2) and (4) don't need to be done at the same time, but if they
-  are then the new release cycle coincides with the old one (up to the
-  renaming of X.0 into X+beta).
-- Coq's X.0 tag can be made as soon as the doc is clear. This will stress the
+  are then the release cycle presented in this CEP roughly coincides with the
+  previous one. Normally (2) precedes (4), but if the doc is not ready yet
+  the platform can still release a beta.
+- Coq's VX.0 tag can be made as soon as the doc is clear. This will stress the
   fact that upstreams can pin with no worries.
 
 ## Announces
 
 - Coq tags are only announced to devs
 - Coq platform pins are communicated to upstream devs, which may tag/release.
-- Coq platform tags and packages are announced to the community
+- Coq platform tags and packages **are announced to the community** (this is
+  when we party)
 
-## Coq CI
+## Coq's CI
 
-- must test standard configurations
-- must test all platform projects with ML parts
-- must test a selection of platform scripts (to test the script themselves,
-  not the build of platform packages)
-- should test platform projects with V files which are depended upon in the
-  platform
+- must test **one configuration**: it cannot use one compiler version or
+  compiler flavor to test one project, and a different one to test another one,
+  since the platform can only have one
+- must test **all platform projects with ML parts**. The platform devs must ensure
+  this invariant, since they are the ones adding packages to the platform
+- must test a **selection of platform scripts** (to test the script themselves,
+  not the build of platform packages). Currently the `-extent=i` platform flag
+  makes it build only Coq+CoqIDE, that should be sufficient.
+- should test **key platform projects** (projects which are heavily depended
+  upon in the platform, even if the are "pure .v projects", e.g. mathcomp).
 - can test anything else of course, including non standard configurations
 
 ## Platform CI
 
-- on branch vX it should test all packages and build installers as artifacts
-- on branch master it should take Coq master and its pins for the subset of
-  packages it covers and test all the packages, eventually report the
-  failures upstream
+- on branch vX it must test all packages and build installers as artifacts.
+  This makes the platform release doable without specific hardware (e.g. a Mac
+  or a PC with windows).
+- on branch master it should take Coq master and its upstream tracked branches
+  (for the subset of projects part of the platform) and eventually report the
+  failures upstream. Currently this activity is logged in dedicated issues.
 
 ### Terminology
 - "severe bug" is a bug which *blocks* many users with no decent workaround,
@@ -147,12 +168,3 @@ This new process identifies 3 groups of developers which need to talk to each
 others: Coq dev, platform devs, and docker devs. This is a risk, but also an
 advantage since the Coq release becomes more lightweight, leaving the RM
 more time to focus on supporting the release with hotfixes.
-
-# Alternatives
-
-We consider docker as *the* platform upstream projects use for testing.
-This may be a bit narrow, even if the docker stuff on coq-community is
-pretty good IMO.
-
-# Unresolved questions
-
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -74,8 +74,8 @@ On time based schedule the RM branches vX.
 
 1. The RM shepherds the few PR which are ready and pins projects tracked by CI
    (using commit hashes, not necessarily tags), then the RM **tags VX+rc**.
-   This step should take no more than 10 days (2 working week).
-   No binary installers are built and published by the RM, just a git tag.
+   This step should take no more than 2 weeks.
+   The RM **tags**, but produces no binary installers, see the platform release.
    The RM writes an **OPAM package**, currently the policy is to upload it to
    `core-dev`.
    Possibly, the RM builds a docker image, so that project maintainers can use it
@@ -83,7 +83,7 @@ On time based schedule the RM branches vX.
    **No breaking change is allowed from now on, unless a severe problem is found.**
 2. The documentation is updated (eg. the Changes file) and eventual fixes
    required by the platform are done. Ideally no other change is done.
-   This step should take no more than 10 days.
+   This step should take no more than 2 weeks.
 3. In response to a severe bug report Coq devs make an hotfix in master which is
    backported to vX by the RM which then tags a point release, possibly as soon
    as the fix is available and merged. The RM writes an OPAM package for
@@ -97,9 +97,9 @@ When Coq VX+rc is tagged, the PRM branches vX
    platform (or its core) are pinned (in accordance with upstreams, which are
    notified about the ongoing process). Most, if not all, packages are in Coq's
    CI or the Platform's CI so there should be no surprises.
-   When done a **VX.0+beta tag** is done and **binary installers** made available to users.
-   A **docker image** with the entire platform prebuilt should be built. This
-   should take 20 days.
+   When done the PRM makes the **VX.0+beta tag** and publishes the **binary installers**.
+   The PRM may also provide a **docker image** with the entire platform prebuilt.
+   This step should not take more than one month.
 5. As users pick up the platform and find severe bugs in Coq, the platform picks
    up point releases of Coq containing hotfixes and eventually extends packages
    beyond the core set.
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -186,7 +186,7 @@ It's essentially the Ubuntu scheme:
 - YYYY.MM plays the role of a major version, as in Coq's 8.13
 - YYYY.MM.0,  YYYY.MM.1 ... play the role of point releases, as Coq's 8.13.0 ..
 
-Hence, it is not YYYY.MM.DD byt rather YYYY.MM.XX.
+Hence, it is not YYYY.MM.DD but rather YYYY.MM.XX.
 
 # Drawbacks
 
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -122,7 +122,7 @@ of the current proposals).
   previous one. Normally (2) precedes (4), but if the doc is not ready yet
   the platform can still release a beta.
 - Coq's VX.0 tag can be made as soon as the doc is clear. This will stress the
-  fact that upstreams can pin with no worries.
+  fact that upstreams can pin with no worries as soon as the rc is tagged.
 
 ## Announces
 
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -72,8 +72,11 @@ Artifacts:
 
 On time based schedule the RM branches vX.
 
-1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed and pins projects tracked by CI
-   (using commit hashes, not necessarily tags).
+1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed
+   and pins projects tracked by CI (using commit hashes, not necessarily tags).
+   The RM generates the Changelog file via the `generate-release-changelog.sh`
+   script, but leaves the release notes empty (it's the project leader which
+   fill them in before the .0 release).
    This step should take no more than 2 weeks.
    - The RM **tags** VX+rc1, but produces no binary installers, see the platform release.
    - The RM writes an **OPAM package**, currently the policy is to upload it to
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -73,21 +73,25 @@ Artifacts:
 On time based schedule the RM branches vX.
 
 1. The RM shepherds the few PR which are ready and pins projects tracked by CI
-   (using commit hashes, not necessarily tags), then the RM **tags VX+rc**.
+   (using commit hashes, not necessarily tags).
    This step should take no more than 2 weeks.
-   The RM **tags**, but produces no binary installers, see the platform release.
-   The RM writes an **OPAM package**, currently the policy is to upload it to
-   `core-dev`.
-   Possibly, the RM builds a docker image, so that project maintainers can use it
-   in CI during (4).
+   - The RM **tags** VX+rc, but produces no binary installers, see the platform release.
+   - The RM writes an **OPAM package**, currently the policy is to upload it to
+     `core-dev`.
+   - Possibly, the RM builds a docker image, so that project maintainers can use it
+     in CI during (4).
+   
    **No breaking change is allowed from now on, unless a severe problem is found.**
 2. The documentation is updated (eg. the Changes file) and eventual fixes
    required by the platform are done. Ideally no other change is done.
    This step should take no more than 2 weeks.
+   - The RM **tags** the .0 release.
+   - The RM writes an **OPAM package** for the main OPAM repository.
 3. In response to a severe bug report Coq devs make an hotfix in master which is
-   backported to vX by the RM which then tags a point release, possibly as soon
-   as the fix is available and merged. The RM writes an OPAM package for
-   the main OPAM repository.
+   backported to vX.
+   - The RM **tags** the point release, possibly as soon
+    as the fix is available and merged.
+   - The RM writes an **OPAM package** for the main OPAM repository.
 
 ## Platform
 
@@ -97,9 +101,9 @@ When Coq VX+rc is tagged, the PRM branches vX
    platform (or its core) are pinned (in accordance with upstreams, which are
    notified about the ongoing process). Most, if not all, packages are in Coq's
    CI or the Platform's CI so there should be no surprises.
-   When done the PRM makes the **VX.0+beta tag** and publishes the **binary installers**.
-   The PRM may also provide a **docker image** with the entire platform prebuilt.
    This step should not take more than one month.
+   - The PRM **tags** VX.0+beta tag and publishes the **binary installers**
+   - The PRM may also provide a **docker image** with the entire platform prebuilt
 5. As users pick up the platform and find severe bugs in Coq, the platform picks
    up point releases of Coq containing hotfixes and eventually extends packages
    beyond the core set.
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -72,7 +72,7 @@ Artifacts:
 
 On time based schedule the RM branches vX.
 
-1. The RM shepherds the few PR which are ready and pins projects tracked by CI
+1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed and pins projects tracked by CI
    (using commit hashes, not necessarily tags).
    This step should take no more than 2 weeks.
    - The RM **tags** VX+rc, but produces no binary installers, see the platform release.
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -136,17 +136,28 @@ of the current proposals).
 
 ## Coq's CI
 
-- must test **one configuration**: it cannot use one compiler version or
-  compiler flavor to test one project, and a different one to test another one,
-  since the platform can only have one
-- must test **all platform projects with ML parts**. The platform devs must ensure
-  this invariant, since they are the ones adding packages to the platform
-- must test a **selection of platform scripts** (to test the script themselves,
-  not the build of platform packages). Currently the `-extent=i` platform flag
+For each project P included in the platform:
+
+- if P has ML code, then it must be tested using the **standard configuration**.
+  It can also be tested using another configuration.
+- if P is pure .v code we have two cases
+  - if it is a core platform component (has many users in the platform) the it
+    must be tested using the standard configuration. It can also be tested using
+    another configuration.
+  - if it is not a core component, then it can be tested with any configuration,
+    or not tested at all (it must be tested by the platform's CI anyway)
+
+The decision of which is the "standard configuration" and the switch to a new
+one is subject to discussion between the Coq developers and the platform ones,
+and the switch should happen in a coordinated way.
+
+Moreover, Coq's CI must test:
+- a **selection of platform scripts**, to test the script themselves,
+  not the build of platform packages. Currently the `-extent=i` platform flag
   makes it build only Coq+CoqIDE, that should be sufficient.
-- should test **key platform projects** (projects which are heavily depended
-  upon in the platform, even if the are "pure .v projects", e.g. mathcomp).
-- can test anything else of course, including non standard configurations
+
+Coq's CI can test anything else of course, including non standard
+configurations and projects that are no in the platform.
 
 ## Platform CI
 
@@ -167,7 +178,9 @@ of the current proposals).
 - "RM" is the release manager of Coq
 - "PRM" are the release managers of the Coq platform
 - "standard configuration" is a version of ocaml and other tools which are
-  agreed between Coq devs and the platform devs (dune, make, gmp, gtk...)
+  agreed between Coq devs and the platform devs (dune, make, gmp, gtk...) as
+  well as a set of configure options or env variables for Coq (eg enabling or
+  disabling native_compute, tuning OCaml's GC, tweaking the stack size)
 
 # Drawbacks
 
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -79,7 +79,7 @@ On time based schedule the RM branches vX.
    fill them in before the .0 release).
    This step should take no more than 2 weeks.
    - The RM **tags** VX+rc1, but produces no binary installers, see the platform release.
-   - The RM writes an **OPAM package**, currently the policy is to upload it to
+   - The RM ensures an **OPAM package** is available, currently the policy is to upload it to
      `core-dev`.
    - Possibly, the RM builds a docker image, so that project maintainers can use it
      in CI during (4).
@@ -89,12 +89,12 @@ On time based schedule the RM branches vX.
    required by the platform are done. Ideally no other change is done.
    This step should take no more than 2 weeks.
    - The RM **tags** the .0 release.
-   - The RM writes an **OPAM package** for the main OPAM repository.
+   - The RM ensures an **OPAM package** is uploaded to the main OPAM repository.
 3. In response to a severe bug report Coq devs make an hotfix in master which is
    backported to vX.
    - The RM **tags** the point release, possibly as soon
     as the fix is available and merged.
-   - The RM writes an **OPAM package** for the main OPAM repository.
+   - The RM ensures an **OPAM package** is uploaded to the main OPAM repository.
 
 ## Platform
 
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -53,7 +53,7 @@ The new process' timeline:
                      DO+           O                         O   O  O
 coq      --+---(1)---+------(2)----+----------(3)------------+---+--+----
  vX branch/          |             |                         |
-           VX+rc tag/              |                         |
+          VX+rc1 tag/              |                         |
                           VX.0 tag/                          |
                                                     VX.1 tag/
    
@@ -75,7 +75,7 @@ On time based schedule the RM branches vX.
 1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed and pins projects tracked by CI
    (using commit hashes, not necessarily tags).
    This step should take no more than 2 weeks.
-   - The RM **tags** VX+rc, but produces no binary installers, see the platform release.
+   - The RM **tags** VX+rc1, but produces no binary installers, see the platform release.
    - The RM writes an **OPAM package**, currently the policy is to upload it to
      `core-dev`.
    - Possibly, the RM builds a docker image, so that project maintainers can use it
@@ -95,7 +95,7 @@ On time based schedule the RM branches vX.
 
 ## Platform
 
-When Coq VX+rc is tagged, the PRM branches vX
+When Coq VX+rc1 is tagged, the PRM branches vX
 
 4. Starting with the pins made by RM on Coq's CI all packages part of the
    platform (or its core) are pinned (in accordance with upstreams, which are
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -50,18 +50,18 @@ Coq is a core component of it.
 
 The new process' timeline:
 ```
-                     DO+           O            O   O  O
-coq      --+---(1)---+------(2)----+-----(3)----+---+--+----
- vX branch/          |             |            |
-           VX+rc tag/              |            |
-                          VX.0 tag/             |
-                                       VX.1 tag/
-
-                                  DI                   DI
-platform  -----------+----(4)------+--------(5)---------+
-           vX branch/              |                    |
-                      VX.0+beta tag/                    |
-                                              VX.0.0 tag/
+                     DO+           O                         O   O  O
+coq      --+---(1)---+------(2)----+----------(3)------------+---+--+----
+ vX branch/          |             |                         |
+           VX+rc tag/              |                         |
+                          VX.0 tag/                          |
+                                                    VX.1 tag/
+   
+                                  DI                   DI      DI
+platform  -----------+----(4)------+--------(5)---------+-------+-------
+           vX branch/              |                    |       |
+                      VX.0+beta tag/                    |       |
+                                              VX.0.0 tag/    .../
 Artifacts:
 - D = docker image for Coq (or Coq + the platform)
 - O = OPAM package (O+ means for core-dev, otherwise it is for the main OPAM repo)
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -63,7 +63,7 @@ platform  -----------+----(4)------+--------(5)---------+-------+-------
                       VX.0+beta tag/                    |       |
                                               VX.0.0 tag/    .../
 Artifacts:
-- D = docker image for Coq (or Coq + the platform)
+- D = Docker image for Coq (or Coq + the platform)
 - O = OPAM package (O+ means for core-dev, otherwise it is for the main OPAM repo)
 - I = binary installers for Coq platform
 ```
@@ -81,7 +81,7 @@ On time based schedule the RM branches vX.
    - The RM **tags** VX+rc1, but produces no binary installers, see the platform release.
    - The RM ensures an **OPAM package** is available, currently the policy is to upload it to
      `core-dev`.
-   - Possibly, the RM builds a docker image, so that project maintainers can use it
+   - Subsequently, the RM ensures a **Docker image** is available, so that project maintainers can use it
      in CI during (4).
    
    **No breaking change is allowed from now on, unless a severe problem is found.**
@@ -106,7 +106,7 @@ When Coq VX+rc1 is tagged, the PRM branches vX
    CI or the Platform's CI so there should be no surprises.
    This step should not take more than one month.
    - The PRM **tags** VX.0+beta tag and publishes the **binary installers**
-   - The PRM may also provide a **docker image** with the entire platform prebuilt
+   - The PRM also ensures a **Docker image** is available with the entire platform prebuilt
 5. As users pick up the platform and find severe bugs in Coq, the platform picks
    up point releases of Coq containing hotfixes and eventually extends packages
    beyond the core set.
@@ -185,6 +185,6 @@ configurations and projects that are no in the platform.
 # Drawbacks
 
 This new process identifies 3 groups of developers which need to talk to each
-others: Coq dev, platform devs, and docker devs. This is a risk, but also an
+others: Coq dev, platform devs, and Docker devs. This is a risk, but also an
 advantage since the Coq release becomes more lightweight, leaving the RM
 more time to focus on supporting the release with hotfixes.
--- a/text/052-platform-release-cycle.md
+++ b/text/052-platform-release-cycle.md
@@ -1,0 +1,158 @@
+- Title: Coq and Coq platform release cycle
+
+- Driver: Enrico
+
+----
+
+# Summary
+
+We aim at a release process for Coq and the Coq platform which makes it
+easier for users to provide feedback and for developers to deliver fixes.
+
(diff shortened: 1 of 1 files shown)


text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+Artifacts:
+- D = docker image for Coq (and opam package)
+- I = binary installers for Coq platform (and opam packages and docker image)
+```
+
+## Coq
+
+On time based schedule the RM branches vX.
+
+1. The RM shepherds the few PR which are ready and pins projects tracked by   CI. Then he tags. This should take 10 days. No packages are built, just a git
```
ejgallego, line 73: he -> they tag

gares: thanks

ejgallego: Also it should be "The RM shepherd" [IMHO it is better to assume everyone is they so we don't have to pick a gender.

ejgallego (approved):
Thanks @gares for the writeup, I think this evolution is on the right path and in line with many of the thoughts that have been discussed over the years.

Indeed the beta should be pretty conservative IMO as you propose, I think this was not so possible before due to actually many regressions happening, but our large (even if expensive) CI has helped a lot to actually produce something that almost works out of the box in most cases.

I'll add some other thoughts inline.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+Coq between the `VX+beta1` and `VX.0` tags. The reasons are, most important
+last:
+- users have better things to do, frankly
+- one month is short, especially with Christmas/summer holydays around
+- in order to test Coq you are likely to need a bunch of Coq packages, and they
+  are typically only available after the beta period is over
+
+Moreover the "beta period" is also seen by devs as a time frame where they can
+still change many things. As a results users are even less happy to spend time
+working on a moving target.
```
ejgallego, line 29: I'm not sure this is true anymore, IMHO I think that most of us really see the beta as a version almost set in stone, but indeed this has been becoming like this gradually.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+
+----
+
+# Summary
+
+We aim at a release process for Coq and the Coq platform which makes it
+easier for users to provide feedback and for developers to deliver fixes.
+
+In a nutshell the Coq "beta period" is replaced by a Coq platform "beta period".
+Coq devs tag a almost final version on top of which the Coq platform core is
```
ejgallego, line 13: an almost

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+
+Moreover the "beta period" is also seen by devs as a time frame where they can
+still change many things. As a results users are even less happy to spend time
+working on a moving target.
+
+Finally, due to the cost of backporting breaking changes from the master
+branch to the release branch, during 8.12 and 8.13 release cycle this only
+happened very early, just after branching (only 2 in the beta period for 8.12,
+0 for 8.13).
+
```
ejgallego, line 35: I'm not sure why this is noted here.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+## New actor: platform
+
+The Coq platform is a selection of Coq package versions which work well
```
ejgallego, line 38: which -> that ?

ejgallego: You could also write just "Coq packages" and to avoid the redundancy with the versions in the next line.

silene:
There are a few points that are not clear to me:
- At which point does pinning happen on the branch? Given the description of stage 4, it seems to have completed at the end of stage 1, but since stage 1 is only 10 days long, that seems unlikely.
- The timeline confuses me a bit. It seems like version X.0 of the platform could be released after version X.1 of Coq. It would seem better to directly skip to platform X.1 in that case, so as to not confuse users.

MSoegtropIMC:
@silene :

A release of a package is mostly there to add features to this package, not to ensure compatibility to core Coq. This made the picking in the past very difficult, because one needed something which does build with latest Coq but does not contain unfinished features of the package. I would say finding a reasonable compromise here was at least 90% of the picking effort in the past - frequently involving putting together cherry picked patches. Since the "rounded up feature set" aspect can now be moved to the Coq platform, I would expect that from now on the picking effort for Coq's CI is much less. For Coq's own CI it doesn't matter if the version tested in CI is something with a reasonably closed feature set one would like to release to the users. All what matters is that whatever is there builds with current Coq. So I think it would be OK to simply pick something which compiles.

Regarding versioning: my plan is that a Coq platform version has always one digit/component more than the Coq release. E.g. there is something like 8.13.0+beta and 8.13.0.0 and likely also a 8.13.0.1. And there might even be something like 8.13+beta.0 or 8.13+beta+beta. Increasing releases of the Coq platform with constant Coq version are e.g. useful to add new packages.

silene:
> So I think it would be OK to simply pick something which compiles.

If I understand you correctly, pinning CI on the Coq branch should not be made using tags (contrarily to what was done on the current release branch) but using random commits. If so, this should certainly be made clearer in the CEP.

gares:
>  If so, this should certainly be made clearer in the CEP.

I will, and mention the (new I guess) script that does this automatically.
FTR there is also another script that opens almost automatically issues on the upstream repo saying "we are testing commit $sha would you mind tagging it"?

gares:
>     * The timeline confuses me a bit. It seems like version X.0 of the platform could be released after version X.1 of Coq. It would seem better to directly skip to platform X.1 in that case, so as to not confuse users.

> Regarding versioning: my plan is that a Coq platform version has always one digit/component more than the Coq release. E.g. there is something like 8.13.0+beta and 8.13.0.0 and likely also a 8.13.0.1. And there might even be something like 8.13+beta.0 or 8.13+beta+beta. Increasing releases of the Coq platform with constant Coq version are e.g. useful to add new packages.

I don't have a clear idea here, nor strong opinions, but I'm not in love with the idea that the platform version mentions the minor version of Coq, especially if we produce many hotfixes. Maybe it's a dream, but it would be nicer if Coq users could say "Coq 8.13 platform 7" to mean the platform version 7 for Coq 8.13 which includes some Coq 8.13.x which the platform devs see fit. After all the plan in this CEP is to hide "bare bone Coq" from users and expose only the platform.

ejgallego:
> I don't have a clear idea here, nor strong opinions, but I'm not in love with the idea that the platform version mentions the minor version of Coq, especially if we produce many hotfixes. Maybe it's a dream, but it would be nicer if Coq users could say "Coq 8.13 platform 7" to mean the platform version 7 for Coq 8.13 which includes some Coq 8.13.x which the platform devs see fit. After all the plan in this CEP is to hide "bare bone Coq" from users and expose only the platform.

A common naming scheme with platforms is to use a YYYY-MM data format, so you have Coq Platform 2020-12 , etc...

text/052-platform-release-cycle.md
Zimmi48, line 73: This could take even less than 10 days (10 days should be an upper bound) if the RM is strict on PR inclusion after the cutoff date.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+- D = docker image for Coq (and opam package)
+- I = binary installers for Coq platform (and opam packages and docker image)
+```
+
+## Coq
+
+On time based schedule the RM branches vX.
+
+1. The RM shepherds the few PR which are ready and pins projects tracked by   CI. Then he tags. This should take 10 days. No packages are built, just a git
+tag. A docker image is build, to that upstreams can use them in CI. No breaking change is allowed from now on.
```
Zimmi48, line 74:
```suggestion
tag. A docker image is build, so that project maintainers can use them in CI. No breaking change is allowed from now on.
```

Zimmi48, line 74: About the "no breaking change allowed", there should be an explicit exception for regression and inconsistency fixes, no?

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+2. Doc is updated (eg. Changes file) and eventual fixes required by the platform
+are done. Ideally no other change is done. This should take 10 days.
+3. In response to a severe bug report Coq devs make an hotfix and the RM tags a point release. Once the fix is produced, a tag comes in 2 days.
```
Zimmi48, line 77: Shouldn't the commitment be with respect to the moment the fix is merged in `master`?

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Platform
+
+When Coq X+rc is tagged, the PRM branches vX
+
+4. Starting with the pins made by RM on Coq's CI all packages part of the platform (or its core) are pinned (in accordance with upstreams, which are notified about the ongoing process). Most, if not all, packages are in Coq's CI so there should be no surprises, otherwise issues are reported to Coq devs. When done a X+rc tag is done and packages made available to users. A docker image with the entire platform prebuilt should be built. This should take 20 days.
```
Zimmi48, line 83: To be consistent with how you called it in your timeline:

```suggestion
4. Starting with the pins made by RM on Coq's CI all packages part of the platform (or its core) are pinned (in accordance with upstreams, which are notified about the ongoing process). Most, if not all, packages are in Coq's CI so there should be no surprises, otherwise issues are reported to Coq devs. When done a X+beta tag is done and packages made available to users. A docker image with the entire platform prebuilt should be built. This should take 20 days.
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,137 @@
...
+5. As users pick up the platforms and find severe bugs in Coq, the platform picks up point releases of Coq containing hotfixes and eventually extends packages beyond the core set.
```
Zimmi48, line 84:
```suggestion
5. As users pick up the platform and find severe bugs in Coq, the platform picks up point releases of Coq containing hotfixes and eventually extends packages beyond the core set.
```

Zimmi48 (approved):
LGTM though I haven't yet read comments from other people.

text/052-platform-release-cycle.md
Zimmi48: Indeed, and this fact is not so clear to users, as some recent discussions about what the beta means have shown.

MSoegtropIMC:
> If I understand you correctly, pinning CI on the Coq branch should not be made using tags (contrarily to what was done on the current release branch) but using random commits. If so, this should certainly be made clearer in the CEP.

Yes, this is the idea as soon as the Windows installer in Coq CI is stripped down to building just Coq and CoqIDE (and maybe Ltac2 and Equations). Up to now using tags was required cause of the Windows installer. But the past experience showed that this approach is hard to maintain and definitely not scalable.

I would say in 8.13 the goal is to make a sort of hybrid approach to have a smooth transition, that is ensure that a Coq Platform with full Windows installer can be build shortly after the release. After that Coq's CI can concentrate on its original purpose.

text/052-platform-release-cycle.md
MSoegtropIMC, line 83: It might make sense to bind the packages to a release branch of the package during the beta phase, rather than to a commit. At the end of the beta phase, this branch is tagged then. Both approaches have their advantages and disadvantages. Using branches is likely more agile and less work for the platform maintainers since changes in the release branches of the packages are immediately visible. The downside is that platform CI does not run automatically on changes, so during the beta phase one would need at least one build a day to find breaking changes. Also with commits the platform maintainers would have more control. I think it depends on the discipline of the package maintainers what works better. My guess is that we will use branches where it works well and fall back to commits for specific projects where things go wild.

Zimmi48:
One thing that is really important IMHO is that this CEP says explicitly what policy it sets and what it leaves up to maintainers / to be decided later. E.g., I wouldn't like the CEP to propose a versioning scheme for the platform to see that it is immediately deviated from, unless the CEP explicitly says that the proposed scheme is just a non-binding example. So, either we explicitly agree on a scheme, details of the release steps, etc., or we explicitly leave some of these items unbound.

My opinion is that it would make sense to have a discussion on the versioning scheme of both Coq and the platform as part of this CEP, but I wouldn't want this to become the whole discussion and to mask other aspects that would deserve to be discussed.

Zimmi48:
> as soon as the Windows installer in Coq CI is stripped down to building just Coq and CoqIDE (and maybe Ltac2 and Equations)

I've already expressed this before and this also seems to be the opinion of Enrico from what I read in this CEP, but I think there should be no left-over code for the Windows installer in Coq's repository and in particular that we should stop producing any kind of installers as assets with Coq's own releases. Of course, keeping a CI job that tests the (external) Coq platform installer scripts makes complete sense.

text/052-platform-release-cycle.md
Zimmi48: > It might make sense to bind the packages to a release branch of the package during the beta phase, rather than to a commit.

Yes, but for most non-plugin projects there are no release branches (releases are created from `master`). So my conclusion is the same as yours.

MSoegtropIMC:
> but I think there should be no left-over code for the Windows installer in Coq's repository and in particular that we should stop producing any kind of installers as assets with Coq's own releases.

I agree. I mentioned the installer just to explain why we did need proper tags before in Coq CI for releases and why this changed now. If one wants to have a small platform build (just Coq and CoqIDE) in Coq CI or not remains to be seen, depending on how frequently it happens that something breaks here. My guess is that it won't be needed. Running a daily master build on Coq platform and manually creating an issue if this breaks should be sufficient.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,158 @@
...
+5. As users pick up the platform and find severe bugs in Coq, the platform picks
+   up point releases of Coq containing hotfixes and eventually extends packages
+   beyond the core set.
+
+## Synchronization points
+
+- The end of (1) starts the release cycle of the platform.
+- The end of (2) and (4) don't need to be done at the same time, but if they
+  are then the new release cycle coincides with the old one (up to the
+  renaming of X.0 into X+beta).
```
erikmd, line 105: Hi all, I've just read the CEP once and after this first reading, it seems the least clear passage is this item.
In particular, I believe one could precise:
* that even if (2) (coq vX.0) and (4) (platform vX+beta) need not to be done at the same time, maybe there should anyway be a constraint such as time(4) ≥ time(2) ?
* and rephrase or elaborate on this sentence: "the new release cycle coincides with the old one (up to the renaming of X.0 into X+beta)" which does not seem clear to me… (w.r.t. "the old one" and "up to the renaming")

erikmd:
Thanks @gares for opening the CEP and pinging me (and sorry for the delay, I've been swamped by teaching, and on-going grading…)
the proposed workflow looks nice; and I strongly agree with [this comment of yours](https://github.com/coq/ceps/pull/52#issuecomment-748616628) where you suggest that within the version `X.0` of the platform, the `X` will just be of the form `8.14` (without making explicit the chosen patchlevel version of coq; even if obviously, the exact version of Coq will be specified and documented using an opam constraint or so).

Indeed, this would be similar to the choice that had been done when I had discussed the versioning strategy of `coqorg/coq` with @Zimmi48 and @ejgallego: we can expect most users of, say, Coq 8.14, to be especially interested in installing Coq 8.14.x with x = the latest released patchlevel, or more generally (in the socpe of this CEP), an appropriate version chosen by the PRM.

Some extra questions:

1. what should happen when Coq V8.14.1 is released and Coq-Platform 8.14.0 already exist? → should we just substitute in the Coq-Platform 8.14.0 spec and docker image, the version of Coq V8.14.0 with Coq V8.14.1? or should we necessarily release a new Coq-Platform 8.14.1, etc., for each change of any version in the spec? (FTR: in `coqorg/coq`, with the current workflow, `coqorg/coq:8.x` is an alias for `coqorg/coq:8.x.0`, and when Coq V8.x.1 is released, `coqorg/coq:8.x.0` is removed and replaced with `coqorg/coq:8.x.1` but the short alias `coqorg/coq:8.x` is still available and is remapped to the latest patchlevel).  
    Actually I guess that for this first question, the answer could be a trade-off: if you think all versions of the libraries should be "immutable" in a given platform version, we might want to have an "exception" for this rule just for the coq compiler patchlevel version itself.

2. as the CEP text proposes to rely on Docker images for packaging the platform, do you think we could directly rely on Docker-Coq images to this aim? (currently, the release of these images follow the opam releases of coq… with the above strategy for packaging the latest patchlevel). In this case, the build of the Coq-Platform images could be similar to building the [`mathcomp/mathcomp`](https://hub.docker.com/r/mathcomp/mathcomp#supported-tags) images. But on second thought, this can be viewed as an "implementation detail" so I guess it doesn't need to be documented in the CEP text, maybe.

3. Currently, there is a difficulty (vicious circle) that just occurs in one place (the CI of Bignums): for backward compatibility reasons, Docker-Coq integrates coq+bignums. But this cause some pain if there is a breaking change in Coq master that breaks Bignums. Because the coq and bignums repository are separate but bundled in the image… which is also used in the Docker-based CI of Bignums, for simplicity. So…  
    Maybe a specific strategy could be devised to have two overlays (in Bigums *and* in Docker-Coq) or, to have a simpler solution, later on, Bignums could just be reintegrated in Coq? (see also https://github.com/coq/opam-coq-archive/pull/1512#issuecomment-744828703; maybe that discussion could be resumed at some point)

gares:
>     2\. But on second thought, this can be viewed as an "implementation detail" so I guess it doesn't need to be documented in the CEP text, maybe.

I think you are right, but before saying that "we officially generate dockers" I wanted a docker expert (eg you) to agree on the plan and eventually help implementing something like the snap job for the platform (so that a PRM does not need to know the details of docker, just how to invoke a job which builds and deploys the image).

gares:
> for backward compatibility reasons, Docker-Coq integrates coq+bignums

I see what you mean. Let me clarify. We have 2 docker images in this CEP:
- one with just coq vX, so that platform packages can test/release (core)
- one with the entire platform, so that all non-platform users can test their stuff against the platform (platform)

To me the core image roughly corresponds to the actual images *but* exists just for technical reasons. In the sense that I would not advertise it to the public in the large, and we may even put in there all we have in Coq's CI (all the plugins, pinned to a commit hash). So yes, also bignums would be there, but not a released version.

The platform image should be the one used for CI in the wild (if you want to CI a project already in there you can start by `opam remove` it). But for a few projects, I believe testing against a stable platform release (as a single entity) should be the preferred solution.

Maybe there is still some need for the current images for both stable and dev Coq, and/or derived images (like the ones with mathcomp.dev).

I start to think docker+CI+platform should be a separate CEP/discussion. Maybe I can just mention the possibility of making a docket image in this CEP.

gares:
I believe I've integrated all remarks and improved the text a little here and there.
I did pick the 8.13.0.0 versioning schema for the sake of the examples, and declared out of the scope of this CEP the discussion about (even if I don't like it, but it is there already, so...).
I did make docker recommended but nor mandatory.

I think this text should be validated one last time by @silene  as the nex RM and by @MSoegtropIMC as the main platform dev.
Once this is done we can try to adjust the release checklist on the Coq side (and maybe craft one on the platform side).

silene:
I am fine with it. Here are a few editorial suggestions. First, since you are counting days using multiple of 5 working days, just write "weeks" instead (e.g., 2 weeks instead of 10 days, 1 month instead of 20 days). Second, I would move all the negated sentences from stage 1 to stage 2; it seems more important to say that stage 2 delivers a binary release rather than stage 1 does not deliver a binary release.

gares:
Thanks I did push some corrections.
I did leave a side note that no binary installers are expected, because it stresses the difference with the current process, and a pointer to the platform release for them. There there was already a positive sentence I guess.

silene:
Perhaps you misunderstood me. Currently, stage 2 reads "The documentation is updated (eg. the Changes file) and eventual fixes required by the platform are done. Ideally no other change is done. [...]" I am saying that you should be more explicit about what the deliverables are. Since there is a `O` in the timeline, I suggest you add "The RM writes an OPAM package and uploads it to the main OPAM package." If you feel it is important to detail the non-delivered things of stage 1, fine with me. But you should detail the yes-delivered things of stage 2.

gares:
Ah, thanks for your patience. It should look better now.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+   - The PRM may also provide a **docker image** with the entire platform prebuilt
+5. As users pick up the platform and find severe bugs in Coq, the platform picks
+   up point releases of Coq containing hotfixes and eventually extends packages
+   beyond the core set.
+
+### Platform versioning
+
+It is out of the scope of this CEP to chose a versioning schema for the
+platform. In the diagram above we append a .digit to the Coq version (one
+of the current proposals).
```
erikmd, line 115: nitpick: this disclaimer looks very OK, but I have one question: assuming "we append a .digit to the Coq version", maybe the bottom rhs tag in the diagram above should be `VX.1.0` instead of `VX.0.0`?

erikmd (approved):
LGTM @gares, thanks!

text/052-platform-release-cycle.md
gares: That's a technical eye! Kind of fixed.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## Coq
+
+On time based schedule the RM branches vX.
+
+1. The RM shepherds the few PR which are ready and pins projects tracked by CI
+   (using commit hashes, not necessarily tags).
+   This step should take no more than 2 weeks.
+   - The RM **tags** VX+rc, but produces no binary installers, see the platform release.
```
Zimmi48, line 78: Use `VX+rc1` and same for the platform beta tag unless somehow the process ensures that there won't be a second rc whatever the situation (and the critical bugs preventing tagging of Coq libraries which are reported).

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+   - The RM writes an **OPAM package**, currently the policy is to upload it to
```
Zimmi48, line 79:
```suggestion
   - The RM writes or request help from opam users to write an **OPAM package**, currently the policy is to upload it to
```

BTW, I probably said that already, but opam is never branded in all caps in the official channels.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+     `core-dev`.
+   - Possibly, the RM builds a docker image, so that project maintainers can use it
```
Zimmi48, line 81:
```suggestion
   - Possibly, the RM builds or request help to build a docker image, so that project maintainers can use it
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+     in CI during (4).
+   
+   **No breaking change is allowed from now on, unless a severe problem is found.**
+2. The documentation is updated (eg. the Changes file) and eventual fixes
```
Zimmi48, line 85: Point 1 should probably explicitly contain a point about minimally updating the changes.rst file by running the automatic consolidation script (leaving it to the .0 release to write the release summary).

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+   required by the platform are done. Ideally no other change is done.
+   This step should take no more than 2 weeks.
+   - The RM **tags** the .0 release.
+   - The RM writes an **OPAM package** for the main OPAM repository.
```
Zimmi48, line 89:
```suggestion
   - The RM writes or request help from opam users to write an **OPAM package** for the main OPAM repository.
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+3. In response to a severe bug report Coq devs make an hotfix in master which is
+   backported to vX.
+   - The RM **tags** the point release, possibly as soon
+    as the fix is available and merged.
+   - The RM writes an **OPAM package** for the main OPAM repository.
```
Zimmi48, line 94:
```suggestion
   - The RM writes or request help from opam users to write an **OPAM package** for the main OPAM repository.
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Synchronization points
+
+- The end of (1) starts the release cycle of the platform.
+- The end of (2) and (4) don't need to be done at the same time, but if they
+  are then the release cycle presented in this CEP roughly coincides with the
+  previous one. Normally (2) precedes (4), but if the doc is not ready yet
+  the platform can still release a beta.
+- Coq's VX.0 tag can be made as soon as the doc is clear. This will stress the
+  fact that upstreams can pin with no worries.
```
Zimmi48, line 125:
```suggestion
  fact that upstreams can pin with no worries as soon as the rc is tagged.
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+- Coq tags are only announced to devs
+- Coq platform pins are communicated to upstream devs, which may tag/release.
+- Coq platform tags and packages **are announced to the community** (this is
+  when we party)
+
+## Coq's CI
+
+- must test **one configuration**: it cannot use one compiler version or
+  compiler flavor to test one project, and a different one to test another one,
+  since the platform can only have one
```
Zimmi48, line 138: This point is unclear and not actionable as is. It seems to be in contradiction with the flexibility left in the last two points below on not testing all platform pure Coq libraries and testing non-standard flavors. It indeed seems to be a good idea that CI tests several configurations (including edge versions of dependencies) so that problems are detected sooner rather than later.
Let's identify current (actual) issues in the Coq CI and fix them (for instance the point below relates to an actual issue).

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,174 @@
...
+- D = docker image for Coq (or Coq + the platform)
+- O = OPAM package (O+ means for core-dev, otherwise it is for the main OPAM repo)
+- I = binary installers for Coq platform
+```
+
+## Coq
+
+On time based schedule the RM branches vX.
+
+1. The RM shepherds the few PR which are ready and pins projects tracked by CI
```
Zimmi48, line 75:
```suggestion
1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed and pins projects tracked by CI
```

Zimmi48 (commented):
The question about the tag name `VX+rc` vs `VX+rc1` is important and has implications. If the former is chosen, it means that there won't be a second rc whatever happens (because version ordering between `VX+rc` and `VX+rc2` won't work as expected). Therefore, it means that if a hotfix is backported to `vX` to allow a specific platform project to tag, the tag won't be compatible with `VX+rc` but only with `X.dev` (until `X.0` is tagged). Therefore, it means that the platform beta won't be doable until `X.0` is tagged (because it can only contain pinned versions). This strategy is OK but should be noted.

text/052-platform-release-cycle.md
gares: I could rephrase it is a more operational way.

For each project P included in the platform:
- if P has ML code, then it must be tested in Coq's CI using the platform configuration. It can also be tested using something else, but not exclusively using another configuration
- if P is pure .v code:
  - if it is a core component (has many users in the platform) the it must be tested in Coq's CI using the platform compiler. It can also be tested using something else, but not exclusively using a non platform compiler.
  - if it is not a core component, then it can be tested with any configuration in Coq's CI, or not tested at all
  - it must be tested by the platform's CI

A configuration is a version/flavor of the OCaml compiler and a set of configure options for Coq/env variables.

The decision of which is the platform configuration and the switch to a new one is subject to discussion between Coq and platform devs, and the switch should happen in a coordinated way.

W.r.t. the problems with Coq's CI, I think (hope) the only one left is about the configuration. IIRC many jobs disable native but I could not reproduce any failure using a standard 4.07.1 compiler on the same gitlab workers. So something tells me we tweak the configuration (I did not investigate yet).

text/052-platform-release-cycle.md
gares: will do, thanks

gares:
I'm not keen in adding "or can ask for help" everywhere. This CEP is not trying to distill the "job profile" of the RM, also there can be (and has been) RM working in tandem, so skills can be shared/complementary.

IMO, you can always ask for help, and you should if you don't know how to proceed. Or, the release doc should be enough detailed to explain you how to do it (eg osx signing).

text/052-platform-release-cycle.md
gares: Is this text more clear? Should I write it this way in the CEP?

text/052-platform-release-cycle.md
gares: done.

text/052-platform-release-cycle.md
Zimmi48: Yes, much clearer. Yes, please update the CEP to be more like what you wrote in this comment.

Zimmi48:
> I'm not keen in adding "or can ask for help" everywhere.

Fine. It's just that up to now, it wasn't the RM job to produce opam packages and Docker images. Maybe you could use a less active language like "ensure opam packages / Docker images are made available" or something like this.

text/052-platform-release-cycle.md
gares: done

Zimmi48 (approved):

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,190 @@
...
+                          VX.0 tag/                          |
+                                                    VX.1 tag/
+   
+                                  DI                   DI      DI
+platform  -----------+----(4)------+--------(5)---------+-------+-------
+           vX branch/              |                    |       |
+                      VX.0+beta tag/                    |       |
+                                              VX.0.0 tag/    .../
+Artifacts:
+- D = docker image for Coq (or Coq + the platform)
```
erikmd, line 66:
```suggestion
- D = Docker image for Coq (or Coq + the platform)
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,190 @@
...
+1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed
+   and pins projects tracked by CI (using commit hashes, not necessarily tags).
+   The RM generates the Changelog file via the `generate-release-changelog.sh`
+   script, but leaves the release notes empty (it's the project leader which
+   fill them in before the .0 release).
+   This step should take no more than 2 weeks.
+   - The RM **tags** VX+rc1, but produces no binary installers, see the platform release.
+   - The RM ensures an **OPAM package** is available, currently the policy is to upload it to
+     `core-dev`.
+   - Possibly, the RM builds a docker image, so that project maintainers can use it
```
erikmd, line 84:
```suggestion
   - Subsequently, the RM ensures a **Docker image** is available, so that project maintainers can use it
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,190 @@
...
+
+When Coq VX+rc1 is tagged, the PRM branches vX
+
+4. Starting with the pins made by RM on Coq's CI all packages part of the
+   platform (or its core) are pinned (in accordance with upstreams, which are
+   notified about the ongoing process). Most, if not all, packages are in Coq's
+   CI or the Platform's CI so there should be no surprises.
+   This step should not take more than one month.
+   - The PRM **tags** VX.0+beta tag and publishes the **binary installers**
+   - The PRM may also provide a **docker image** with the entire platform prebuilt
```
erikmd, line 109:
```suggestion
   - The PRM also ensures a **Docker image** is available with the entire platform prebuilt
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,190 @@
...
+- "PRM" are the release managers of the Coq platform
+- "standard configuration" is a version of ocaml and other tools which are
+  agreed between Coq devs and the platform devs (dune, make, gmp, gtk...) as
+  well as a set of configure options or env variables for Coq (eg enabling or
+  disabling native_compute, tuning OCaml's GC, tweaking the stack size)
+
+# Drawbacks
+
+This new process identifies 3 groups of developers which need to talk to each
+others: Coq dev, platform devs, and docker devs. This is a risk, but also an
```
erikmd, line 188:
```suggestion
others: Coq dev, platform devs, and Docker devs. This is a risk, but also an
```

erikmd (approved):
Thanks @Zimmi48 and @gares for further enhancing the text!

I skimmed the CEP another time; I left a few suggestions (cf. the inline comments), to sum up:
* capitalizing the "Docker" word (while `docker` would rather be the CLI program spelling)
* adopting the same prose for Docker and OPAM sentences:  
  In particular the words "Possibly", "may also provide" were not that precise, and maybe unneeded: I don't believe you wanted to say that "the build of the Docker images is optional" (given e.g. [that line](https://github.com/coq/ceps/pull/52/commits/fe0e7379bee8b7035f4831a2643db2c1f3234e77#diff-a99a297e024575e11ff21dca755e2ecceb42e2d4a4ec4bce91625a35767dad4dR153) that was part of the first commit of the CEP), but maybe the fact that "the RM/PRM just needs to ensure the Docker release is done": either by posting a request in a GitHub issue or so (like you did for Coq 8.13), or by directly looking after the build (opening a PR to the CI/CD related repository).

gares:
Thanks

erikmd:
Hi all, I know that it has been mentioned above that the precise naming convention could be discussed in a later thread.

However I just wanted to raise a suggestion on a detail (that might impact the suffix that is mentioned in the CEP):

as the Coq platform can be seen as a **distribution** of Coq (core + contributions), would it make sense to always append the suffix with a `+` instead of just a `.`? a bit like how some **GNU/Linux distributions** such as Debian tag the packaged versions… I mean, that would amount to release for Coq 8.14.0:
* coq-platform 8.14.0+beta; coq-platform 8.14.0+0; coq-platform 8.14.0+1 (assuming we keep Coq's whole 3-place version)
* or coq-platform 8.14+beta; coq-platform 8.14+0; coq-platform 8.14+1 (assuming we want to "hide" Coq's point release to users, while avoiding any ambiguity w.r.t. what "8.14.1" should mean).

WDYT?

gares:
Your second option would be my choice.

Zimmi48:
Thanks for this suggestion @erikmd. I would support it (especially the second choice). I'll repeat it in the appropriate thread.

MSoegtropIMC (approved):
I reread the document from beginning to end again and it looks good to me. It well describes and details the process I had in mind when starting the Coq platform project.

erikmd:
Hi @Zimmi48 

> Thanks for this suggestion

The only doubt I have actually, is related to the implied order:

If `x+beta < x+0` then we're very OK :smiley:

gares:
Well spot!

```
gares@ollypat:~/COQ/backports-8.13$ dpkg --compare-versions 1.1+beta1 lt 1.1+0 ; echo $?
1
```

We have 2 alternatives:
```
gares@ollypat:~/COQ/backports-8.13$ dpkg --compare-versions 1.1beta1 lt 1.1+0 ; echo $?
0
gares@ollypat:~/COQ/backports-8.13$ dpkg --compare-versions 1.1~beta1 lt 1.1+0 ; echo $?
0
```

gares:
But IIRC `~` is not OK in git's tags, and nix also dislikes it

gares:
(or we stop calling it beta, and we start from `+0`)

Zimmi48:
> (or we stop calling it beta, and we start from `+0`)

Indeed, but then, I would add one more digit:
- `X+0.1` / `X+0.2` (instead of `X+beta1` / `X+beta2`)
- `X+1.0` / `X+1.1` (instead of `X+1` / `X+2`)
- `X+2.0` in the exceptional case where a new platform release ships with more than bug fix updates, such as when adding a new package to the platform.

Reaching `+1.0` would indicate that the platform is out of the testing phase.

MSoegtropIMC:
Or we go back to:
```
8.14+beta
8.14.0
```
as we have in Coq. There is some chance for confusing this with Coq patch levels, but as discussed if the Coq platform is the officially delivered product, the awareness for the Coq patch level might fade quickly and only experts will know what this is.

erikmd:
Hi, but if we fallback to `8.14+beta` - `8.14.0`, shouldn't the beta have a numeric suffix as well? (`8.14+beta1`)

MSoegtropIMC:
> Hi, but if we fallback to 8.14+beta - 8.14.0, shouldn't the beta have a numeric suffix as well? (8.14+beta1)

Yes, indeed.

gares:
The more I think about it and the more I smell confusion. The problem is that 8.14.0 will be both a Coq tag and a Platform tag/release, which seems like looking for troubles.

Maybe we should reconsider @ejgallego proposal of using a date, e.g. 2021.06+beta1 and 2021.06.x, and give up injecting the major version of Coq inside the Platform version. It would also a major platform release without a Coq one, eg 2021.09.0 still based on Coq 8.14 release in June.

silene:
I did not want to comment here, since this discussion was supposed to happen elsewhere. But since this did not happen, let met say I am in agreement with @ejgallego and we should use a versioning scheme à la Ubuntu.

MSoegtropIMC:
@gares, @silene : I have issues with some open source projects I use which have date tags - it is sometimes hard to tell if something is a minor improvement or a more radical change. For Coq with its well managed release cycle this would not be an issue, though. It would make sense to change this for Coq as well. Otherwise we will be at 8.23 in 5 years.

So yes, definitely also a reasonable option.

silene:
> It would make sense to change this for Coq as well.

I do not agree. A release of the platform is a curated snapshot of the Coq ecosystem at a given point in time. The fact that we intend to more or less align these releases with those of Coq does not change that fact.

For Coq, this is quite different, as it follows the semantic versioning principles, although Coq developers have rather diverging opinions on what changes are major or minor. (At the time of Coq 8.5, I was advocating that it should be released as 9.0. I was quite alone in that fight.)

SkySkimmer:
>For Coq, this is quite different, as it follows the semantic versioning principles

Strong disagree, we do breaking changes every release even though we only change the "minor" digit.

MSoegtropIMC:
> For Coq, this is quite different, as it follows the semantic versioning principles

I don't think this matches the current release policy, which is more or less to make one release every 6 months. So if the release policy is time based and not semantics based, it would make sense to reflect this in the version numbering scheme.

ejgallego:
Indeed it seems to me that a platform such as the one we have for Coq is too complex as to be assigned a single version number so indeed a date-based numbering is quite informative for users.

The issue of "Coq language" vs "Coq implementation" has been indeed lingering for a while, it makes sense that at some point we could declare a bump in the "Coq language" , so you would have something such as "Coq platform 2021.04 supporting Coq language versions 8.12.x" .

Coq language versions could indeed be supported semantically, and even at some point we could require users to declare the Coq language versions they are using. Big (huge) problem with this is that supporting more than one language version in an implementation is a lot of effort, so I dunno.

gares:
I don't have a strong opinion about using a date for Coq as well, but we can kind of follow [semver](https://semver.org/) for Coq, or at least try. So maybe we have something to lose by moving away from the 3 digit schema.

Semver for the platform seems hopeless, hence I don't mind drifting away from a semver-like schema immediately.

PS: Coq 9.x will remove `.` from the end of sentences, STM+univpoly was a piece of cake!

Zimmi48:
> PS: Coq 9.x will remove `.` from the end of sentences

LOL certainly not.

Anyway, I'm forced to comment to repeat @SkySkimmer's comment that we *do not* at that time follow semantic versioning since each 8.X version is a major release according to the semantic versioning definition (i.e., they come with breaking changes). We do not produce any minor releases (only major and patch-level releases). I've been considering for a while the idea of switching to a true semantic versioning release cycle for Coq (i.e., release minor versions with new features but zero breaking changes say every 4 months and major versions with breaking changes only say every year) but every person I've discussed this idea with has expressed that this might not apply well to the evolution of a language.

erikmd:
@ejgallego 

> The issue of "Coq language" vs "Coq implementation" has been indeed lingering for a while, it makes sense that at some point we could declare a bump in the "Coq language" , so you would have something such as "Coq platform 2021.04 supporting Coq language versions 8.12.x" .

Looks good.

> Coq language versions could indeed be supported semantically, and even at some point we could require users to declare the Coq language versions they are using.

Indeed… a bit like dune-project's `(lang dune 2.8)` spec.

> Big (huge) problem with this is that supporting more than one language version in an implementation is a lot of effort, so I dunno.

Agree. All the more as "supporting" means keeping binaries / images for all supported versions in a single repo / registry.

> Indeed it seems to me that a platform such as the one we have for Coq is too complex as to be assigned a single version number so indeed a date-based numbering is quite informative for users.

FWIW, there are some version numbering schemes that combine both ideas (namely, partial semver + some [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)-like date):
https://git.joeyh.name/index.cgi/git-annex.git/refs/
> 8.20210127
> 8.20201129
> 8.20201127
> 8.20201116
> 8.20201103
> 8.20201007
> 8.20200908
> 8.20200810
> 8.20200720.1
> …

so in this perspective, one could think of `Coq Platform 8.14.YYYYMMDD` for example (even if it's a bit verbose…)

Zimmi48:
Note that the date-only solution doesn't solve the version ordering issue of `YYYYMMDD` vs `YYYYMMDD+beta1` (although it would be unexpected that the beta and the final release are released on the same date but then it means that beta are associated to no "final release").

The Coq version + date solution of @erikmd just above has the advantage that it still allows you to release a `8.14+beta1` (and if needed a `8.14+beta2`) before a final `8.14+YYYYMMDD`.

silene:
Note that @ejgallego was not suggesting a complete date and neither I am.  Having just `21.02.0` (year/month of the original release + bugfix) is fine with me. (This is the versioning scheme of Ubuntu.) Also, with such a naming scheme, there is just no issue with beta versions: `21.02+beta` is smaller than `21.02.0` which is smaller than `21.02.1`.

MSoegtropIMC:
One note on date based solutions: this might imply that all packages are the latest release versions of that date. This conflicts a bit with the idea of keeping the versions of the delivered packages constant for a while - the initial thought was during one Coq release cycle. E.g. I planned to add new packages to the Coq platform v8.13 branch, but not to change the versions of existing packages until 8.14.

Is there a Coq version number based scheme with in total 3 numbers or two numbers and beta<number> which is compatible with all system version comparison rules and is *not* 8.13+beta0 / 8.13.0? The latter was rejected cause of possible confusion with Coq versions.

silene:
> One note on date based solutions: this might imply that all packages are the latest release versions of that date. This conflicts a bit with the idea of keeping the versions of the delivered packages constant for a while

Note that, with the `21.02.x` style of date, the first two numbers do not change. So, most packages can stay at the same version. For example, Ubuntu 20.04.2 will be released in February 2021, but users do not expect it to contain packages that were released before April 2020. If they want more recent packages, they should use version 20.10.

gares:
>  this might imply that all packages are the latest release versions of that date.

At least in the Debian/Ubuntu tradition this is not the case. Eg 2020.04.0 does not hold the last available version of any software it ships on that date; 2020.04.1 may hold software that was patched (or even released in case of the kernel) in 2021. In some sense the date is just a number, encoded in a way which makes you quickly see if it is old or new.

Zimmi48:
> Is there a Coq version number based scheme with in total 3 numbers or two numbers and beta which is compatible with all system version comparison rules and is _not_ 8.13+beta0 / 8.13.0? The latter was rejected cause of possible confusion with Coq versions.

You put a lot of constraints here. With only three numbers but the Coq major version in it, this means that two numbers are taken by Coq. This doesn't leave out many options. One is to use `YYYYMMDD` (or `YYMMDD`) for the last number.

Whereas if we do accept to not include the Coq major version in the platform version, then the solution proposed by @silene above (`21.02.X`) is a good one.

Zimmi48:
It seems that we have converged on the `YYYY.MM.X` scheme for the platform versions (see https://coq.zulipchat.com/#narrow/stream/250632-Coq-Platform.20devs.20.26.20users/topic/Version.20naming). I guess that since we have arrived to this conclusion, it would make sense to update the CEP accordingly. Then, it seems to me that it is time to merge.

Note that I've updated the release checklist accordingly in coq/coq#13823.

gares:
I'll try to update the CEP tomorrow, thanks everybody for contributing to the discussion

gares:
I hope my changes are OK w.r.t. the version of the platform.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+----
+
+# Summary
+
+We aim at a release process for Coq and the Coq platform which makes it
+easier for users to provide feedback and for developers to deliver fixes.
+
+In a nutshell the *Coq "beta period"* is replaced by a **Coq platform "beta period"**.
+Coq devs tag an almost final version on top of which the Coq platform core is
+built. Coq point release are then used to quickly deliver fixes
```
Zimmi48, line 14:
```suggestion
built. Coq point releases are then used to quickly deliver fixes
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+to the platform.
+
+# Motivation
+
+The beta testing period does not provide enough feedback to sensibly improve
+Coq between the `VX+beta1` and `VX.0` tags. The reasons are, most important
+last:
+- users have better things to do, frankly
+- one month is short, especially with Christmas, summer holydays or conferences
```
Zimmi48, line 23:
```suggestion
- one month is short, especially with Christmas, summer holidays or conferences
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+  around (our beta period is December or May)
+- in order to test Coq you are likely to need a bunch of Coq packages, and they
+  are typically only available after the beta period is over
+
+Moreover the "beta period" has also been seen by devs as a time frame where
+they can still change many things. As a results users are even less happy
```
Zimmi48, line 29:
```suggestion
they can still change many things. As a result, users are even less happy
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+- D = Docker image for Coq (or Coq + the platform)
+- O = OPAM package (O+ means for core-dev, otherwise it is for the main OPAM repo)
+- I = binary installers for Coq platform
+```
+
+## Coq
+
+On time based schedule the RM branches vX.
+
+1. The RM shepherds the few PR which are ready, ensures blocker issues are fixed
```
Zimmi48, line 75:
```suggestion
1. The RM shepherds the few PRs which are ready, ensures blocker issues are fixed
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+   and pins projects tracked by CI (using commit hashes, not necessarily tags).
+   The RM generates the Changelog file via the `generate-release-changelog.sh`
+   script, but leaves the release notes empty (it's the project leader which
+   fill them in before the .0 release).
```
Zimmi48, line 79:
```suggestion
   script, but leaves the release notes empty (it's the project leader who
   fills them in before the .0 release).
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+   - The PRM also ensures a **Docker image** is available with the entire platform prebuilt
+5. As users pick up the platform and find severe bugs in Coq, the platform picks
+   up point releases of Coq containing hotfixes and eventually extends packages
+   beyond the core set.
+
+### Platform versioning
+
+It is out of the scope of this CEP to chose a versioning schema for the
+platform. In the diagram above we append a .digit to the Coq version (one
+of the current proposals).
```
Zimmi48, line 118: This is outdated. It can probably be simply removed:
```suggestion
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+  when we party)
+
+## Coq's CI
+
+For each project P included in the platform:
+
+- if P has ML code, then it must be tested using the **standard configuration**.
+  It can also be tested using another configuration.
+- if P is pure .v code we have two cases
+  - if it is a core platform component (has many users in the platform) the it
```
Zimmi48, line 144:
```suggestion
  - if it is a core platform component (has many users in the platform) then it
```

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,201 @@
...
+one is subject to discussion between the Coq developers and the platform ones,
+and the switch should happen in a coordinated way.
+
+Moreover, Coq's CI must test:
+- a **selection of platform scripts**, to test the script themselves,
+  not the build of platform packages. Currently the `-extent=i` platform flag
+  makes it build only Coq+CoqIDE, that should be sufficient.
+
+Coq's CI can test anything else of course, including non standard
+configurations and projects that are no in the platform.
```
Zimmi48, line 160:
```suggestion
configurations and projects that are not in the platform.
```

Zimmi48 (commented):
These are only typos.

gares:
Thanks! and agreed.

text/052-platform-release-cycle.md
```diff
@@ -0,0 +1,196 @@
...
+  disabling native_compute, tuning OCaml's GC, tweaking the stack size)
+
+
+### Notes about platform versioning
+
+It's essentially the Ubuntu scheme:
+- YYYY.MM plays the role of a major version, as in Coq's 8.13
+- YYYY.MM.0,  YYYY.MM.1 ... play the role of point releases, as Coq's 8.13.0 ..
+
+Hence, it is not YYYY.MM.DD byt rather YYYY.MM.XX.
```
gares, line 189:
```suggestion
Hence, it is not YYYY.MM.DD but rather YYYY.MM.XX.
```

gares:
we can merge now

Zimmi48:
Approved in the Coq Call :wink:

Zimmi48:
I've renamed the two impacted milestones in the Coq repository.

erikmd:
@gares, I've just noticed that given the underlying feature branch of this PR does not exist anymore, the URL
`https://github.com/gares/ceps/blob/platform-release-cycle/text/052-platform-release-cycle.md` in https://github.com/coq/ceps/pull/52#issue-771409085 is broken… so you may want to edit the initial post to put https://github.com/coq/ceps/blob/master/text/052-platform-release-cycle.md

PR #54, opened 2021-01-25
[RFC] limited conditional compilation

gares:
Use case mock ups:

- compat
```coq
#[if(COQ <= "8.13")] Old Command.
#[if("8.13" < COQ)] New Command.
```

- code gen
```coq
#[if(derive)] Require Derive.

#generated(derive).
  (* does not need Derive in order to execute *)
  Module bar_utils.
    Definition foo := ....
    Inductive baz := ....
  End bar_utils.
#with.
  (* implemented by plugin Derive, does not even parse without Derive *)
  Derive foo, baz For bar.
#end.
```

- code share
```coq
#[if(ARCH = "32")] From Compcert32 Require Import X.
#[if(ARCH = "64")] From Compcert64 Require Import X.
```

```coq
#[if(TEST)] Check this type checks.
#[if(TEST)] Check (erefl :> program input = output).
```

### DIFF
--- a/text/054-conditional-comments.md
+++ b/text/054-conditional-comments.md
@@ -1,0 +1,209 @@
+- Title: Conditional Compilation
+
+- Drivers: Enrico Tassi
+
+----
+
+# Summary
+
+Most programming languages provide out of the box support for some form
+of conditional compilation, either via a preprocessor or by integrating
+in the syntax conditionally compiled blocks.
+
+This CEP proposes a similar feature for Coq documents, but constraining it
+in order to make it easy to integrate it in existing tools and not mess up
+documents too much.
+
+# Motivation
+
+I see 3 main use cases:
+- backward compatibility
+- code generation
+- code sharing
+
+### Backward compatibility
+
+One common use case for conditional compilation is to allow the same source
+file to work in two incompatible versions of the language
+
+```coq
+#[if(COQ <= "8.13")] Old Command.
+#[if("8.13" < COQ)] New Command.
+```
+
+This would make it always possible to write backward compatible overlays for vernac files. One could also imagine a warning the user would enable when she decides to drop compatibility with a Coq version. If any condition is true (assuming a Coq version to drop) then the warning is raised and the user removed the line or tightens the bound.
+
+### Code generation
+
+One problem about code generation this CEP tackles is the need to have the
+generator around in order to check the document (or having the generator take
+over the entire file).
+
+```coq
+#[if(derive)] Require Derive.
+
+#generated(derive).
+  (* does not need Derive in order to execute *)
+  Module bar_utils.
+    Definition foo := ....
+    Inductive baz := ....
+  End bar_utils.
+#with.
+  (* implemented by plugin Derive, does not even parse without Derive *)
+  Derive foo, baz For bar.
+#end.
+```
+
+If `derive` is not defined, the generated code is used,
+otherwise the generator code is run (and its output is recorded and 
+eventually replaces the old code, not in this CEP, which only provides a
+place to store the output and conditionally run the generator or the
+generated code).
+
+This helps when `Derive` is a project with many dependencies one does not
+want to impose to users. One cannot use `#[if(derive)] Derive...` since that
+sentence would not even parse without the generator around.
+
+Another example inspired by tactician:
+
+```coq
+Lemma foo : statement.
+Proof.
+#generated(fancy).
+  simple_tactic_using hint1 hint2.
+#with.
+  fancy_tac. (* takes time, require ML stuff around *)
+#end.
+Qed.
+```
+
+UIs hide/fold the generated code, which should then display as:
+
+```coq
+#[if(fancy)] Require FancyTactic.
+
+Lemma foo : statement.
+Proof.
+▷ fancy_tac.
+Qed.
+```
+
+and unfold to the former when `▷` is clicked. 
+
+### Code sharing
+
+Some users raised the need for running the same fine against a different
+set of imports, e.g. the architecture for which one wants to verify a file.
+
+```coq
+#[if(ARCH = "32")] From Compcert32 Require Import X.
+#[if(ARCH = "64")] From Compcert64 Require Import X.
+```
+
+# Detailed design
+
+The design makes the vernacular command interpreter aware of these attributes
+and execute as no-op code which is not selected. In order to make this task
+reasonably easy we impose the following limitations.
+
+A common piece of code interpreting the conditional expressions is placed in
+`lib/` and shared by all components.
+
+## Conditions
+
+The grammar
+
+```
+COND ::= VERSION_TEST | VARIABLE_TEST
+VERSION_TEST ::= PRODUCT UBOUND | LBOUND PRODUCT | LBOUND PRODUCT UBOUND
+PRODUCT ::= COQ | <ident>
+UBOUND ::= < "VERSION" | <= "VERSION"
+LBOUND ::= "VERSION" < | "VERSION" <=
+VERSION ::= <num> | <num> . <num> | <num> . <num> . <num>
+VARIABLE_TEST ::= <ident> | <ident> = <string>
+```
+
+Examples
+```
+"8.13" < COQ < "8.14.2" (* from 8.13+alpha to 1.14.1 *)
+"8.13.0" <= COQ (* to rule out +something *)
+"1.4" <= EQUATIONS (* 1.4 ... 1.4+8.13 ... 1.4+8.14 *)
+FOO
+FOO = "yes"
+FOO = "99"
+```
+
+Version comparisons as per OPAM, but with a very simplified grammar,
+only 1 to 3 digits in bounds, `+something` is epsilon.
+
+## No nesting
+
+`#generated.` ... `#with.` `#end.` blocks cannot nest.
+
+This limitation is not imposed by an implementation difficulty, but rather to keep files understandable.
+
+## No `Require` in generated code
+
+The `Require` statement cannot occur in a generated block, in order to
+simplify the job of `coqdep` (see below).
+
+## Condition variables are immutable
+
+Tools `coqc`, `coq*top` and `coqdep` accept `-D VAR=val` and there is no other
+way to set variables used in the conditionals. There is no `Set Variable Foo.`
+no `#define FOO.` and no `getenv("FOO")`.
+
+`coq_makefile` and Dune's `coq-lang` can use their own ways to pass options to
+the tools.
+
+`coqdep` evaluates the conditional expression in from of `Require` statements
+and spits the dependency if needed. This is enough to cover `coq_makefile` and
+`dune`. 
+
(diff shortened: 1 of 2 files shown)


silene:
It seems to me like `#generated` and `#with` are just fancy names for `#if` and `#else`. What is the rationale for not going with the universally recognized names?

By the way, I am a heavy users of compatibility lemmas. For example, the `src/Missing` directory of CoqInterval contains 2500 lines of lemmas. I would like to make them conditional on the version of the corresponding libraries. This CEP does not help much in this regard.

On a related note, it is unfortunate that `#define` is not provided. Currently, Flocq contains
```coq
Definition Flocq_version := Eval vm_compute in
  let fix parse s major minor :=
    match s with
    | String "."%char t => parse t (major * 100 + minor)%N N0
    | String h t =>
      parse t major (minor * 10 + N_of_ascii h - N_of_ascii "0"%char)%N
    | Empty_string => (major * 100 + minor)%N
    end in
  parse "3.4.0"%string N0 N0.
```

I would have liked to be able to just write `#define Flocq_version "3.4.0"` instead.

In case you wonder, the `configure` script from CoqInterval compiles the following code to check the version of Flocq.
```coq
Require Import Flocq.Version BinNat.
Goal (30100 <= Flocq_version)%N. easy. Qed.
```

gares:
> What is the rationale for not going with the universally recognized names?

The use case about generated code. I'd like this "if-then-else" to be special and be handled by other "tools". In particular the text between `#generated` and `#with` is to be replaced automatically when running the text between `#with` and `#end`. This is not written in this CEP (just alluded to, it's called promotion in dune).

gares:
> I would have liked to be able to just write `#define Flocq_version "3.4.0"` instead.

Ah ok, this makes sense. IMO the best would be to let each project define a key/value for `PRODUCT` (in the grammar). This would be a very weak form of `#define` (it would not be possible to write that in the middle of a file, that is it's value is always there).

gares:
> CoqInterval contains 2500 lines of lemmas. I would like to make them conditional on the version of the corresponding libraries. This CEP does not help much in this regard.

You mean that you would need an `#if-#then-#end` for that?

silene:
> I'd like this "if-then-else" to be special and be handled by other "tools".

I see. But that still seems like a weak reason for having a brand new syntax. The exact same code could have used `#if` / `#else` with the exact same effect. If an external tool is able to look for `#generated(derive)` and modify the next few lines of a `.v` file, then it is also able to look for `#if(derive)` and modify the next few lines of a `.v` file.

> You mean that you would need an `#if(). ... #end.` for that? (if so, I'm fine with this)

It is not really needed. I have lived without it till now. But it would certainly be helpful. There is no reason to provide forward compatible lemmas if the ones provided by the imported libraries already have the correct statements.

In an ideal world, I would like to get rid of the `SpecFloatCompat.v.in` file of Flocq which contains a single line:
```coq
Require Export @SPEC_FLOAT_MODULE@.
```

(And yes, the `@SPEC_FLOAT_MODULE@` part is substituted by the `configure` script, depending on the version of Coq.)

By the way, you seem worried about the interaction between `coqdep` and conditional compilation. I don't think it matters much. For example, I am assuming that, in your use case with `Derive`, the library is actually external. So, `coqdep` will just ignore this dependency anyway. So, it does not need to know about `#[if(derive)]`. The same is true of your CompCert use case. As a side note, we do not even need `#define` if `coqdep` is oblivious to macros. A plain `Definition` would be enough to trigger conditional compilation.

The issue with `coqdep` becomes a bit more subtle when it comes to files internal to a library. My Flocq use case with `@SPEC_FLOAT_MODULE@` above would roughly translate to:
```coq
#[if(COQ <  "8.11")] Require Import SpecFloatCopy.
#[if(COQ >= "8.11")] From Coq Require Import SpecFloat.
```

But actually, it would still work fine if `coqdep` were to put a spurious dependency on `SpecFloatCopy`. Sure, this would cause `SpecFloatCopy.vo` to be compiled even for Coq >= 8.11. But it is just a waste of time, nothing more.

gares:
> The exact same code could have used `#if` / `#else` with the exact same effect.

Right. But UIs would not know it's a good idea to fold away the then branch.
Anyway, let's see what others think. I'm OK having just if-then-else or both.

gares:
> Sure, this would cause `SpecFloatCopy.vo` to be compiled even for Coq >= 8.11. But it is just a waste of time, nothing more.

It seems better to spit out the right dependencies rather than hoping that an over approximation works.

I quite don't get what is your point here. Do you have a use case which is ruled out by the restrictions I did put in place?
(many can be relaxed later, I'd rather be conservative on a first proposal).

silene:
> I quite don't get what is your point here.

What I mean is that most of the restrictions you put on your system seem to come from the fact that you want `coqdep` to be able to select the correct conditional branches. This severely cripples the proposal. If you remove this bottleneck, then a lot more becomes possible, without making the code of `coqtop` more complicated. Example:

```coq
#if not defined(foo).
Lemma foo : ...
Proof. ... Qed.
#end.
```

Note that the conditional now refers to the content of a `.vo` file, which `coqdep` has no hope of ever being able to access.

gares:
Well, I try to add a very constrained `cpp` and you propose something `cpp` could not even do, since the test is about a runtime property. I now understand you point of view, but I guess I'll wait to see in which direction others lean as well.

ejgallego:
Thanks @gares for the detailed writeup, a few comments:
- indeed I do remain worried about the choice of having the vernacular interpretation be responsible of the implementation of this kind of transformation. IMHO that is a choice both problematic from:
  + the architectural point of view [as indeed fully delegates to runtime key things such as constant declaration, think for example that now generating the document outline would require some form of execution, but that problem also persists when using a preprocessor]
  + the scoping point of view, as for example general `#if #then #else` became hard to do, also I don't understand how the `derive` declarations would work here now if they are gonna span more than one sentence.
- IMHO we should define a bit more structurally what the scope of this attributes are, in terms of transformations. For example, most PPX in OCaml are defined as AST to AST combinators, however others actually transform the typed tree. In Coq you also have the runtime level. It seems to me that both the current proposal and @silene suggestion go pretty close to the runtime level.
- going to the runtime level is IMO a bit different from conditional compilation. Similar features have been long requested for Coq, under the name of "vernacular scripting" [so for example we can have a way to generate lemmas, etc...] IMHO that's a very interesting feature, but pretty ambitious and way beyond what is needed for compatibility.
- I don't think that having the build system take care of conditional compilation is out of the question, already in the OCaml land you have examples that use the build system to fix inline tests, and you can have rules to generate compat files easily, with `(cp foo.%COQ_VERSION foo.v)` , etc... Moreover, the build system has a better change to know for example versions of libraries on which a project depends on.
- Experience with other systems seem to indicate that Coq would indeed tend to integrate the build system into the compiler anyways, as coupling tends to be quite tight and incremental compilation requires it, however that is also a bit far into the future.

So overall, I am still quite wary about the implementation design, in particular about hooking the attributes to the runtime interpretation of commands. Using a simple pre-processor or defining a layer similar to ppx for doing some kind of pre-processing / meta-programming could be a better solution in the medium term I think.

gares:
I don't see how using an external preprocessor is easy. Eg ppx in ocaml are messy for UIs because they have to tell the language server that the text they are sending has to be processed with some options that come from the build system. Indeed now you need the right version of dune, ocamllsp and ocaml-platform in order to comfortably edit an ocaml file which uses a ppx. As a user, it is a pain point.

What I'm trying to do here is to avoid using an external tool and rather integrate in Coq a minimalistic preprocessor covering some use cases. That should ease deployment, at the cost of not covering all possible usecases.

Yes, it would be Coq to decide if a sentence is to be executed or not, you are right (but it is not really execution as in vernacinterp, the conditionals can all be evaluated immediately given the restrictions I currently put). Having external tools do something sensible without asking Coq is what motivated my choice of using `#generated` rather than `#if`, since in this way you always know what to display and what not (also the rules banning nesting or non-well-parenthesized code are there to help with this).

ejgallego:
> What I'm trying to do here is to avoid using an external tool and rather integrate in Coq a minimalistic preprocessor covering some use cases. That should ease deployment, at the cost of not covering all possible usecases.

It is fine we just provide one preprocessor in Coq, as to avoid the problems [I also believe that OCaml should have many of the ppx as standand in the compiler] A bit of my point was that for example, to have a few `#ifdef` we could just use cpp and save some time, maybe indeed I am hoping that Coq documents use very seldom this feature and don't become an if-def mess. Frankly for that use case I'd rather have users to rely on the build system and a few aux files.

But that was not my core worry but more of a comment; I am fine having such a pre-processor, but my main question is where does the pre-processor hook up, if I understand currently you are not proposing to add a pre-processor but to modify the core processing routine, this is what worries me and has the one-sentence scoping problem, but likely I'm biased here by the implementation provided in the PR.

Can you detail a bit more how the data-flow of document parsing / evaluation would work here?

gares:
There are two cases: single sentence and multiple.

- Sentences like `#[if(cond)] Sentence.` are parsed as usual, classified/typed as no-op if cond is false, executed as such. Eg. `#[if(false)] Print nat.` would be a regular sentence, it is just that it won't print anything.

Let me recall the other case
```
#generaded(cond).
  Definition foo := ...
  Definition bat := ...
#with.
  Derive foo, bar for baz.
#end.
```

- Sentence/floating-attribute `#generated(cond).` parses as a regular sentence, typed as VtDefault. It's execution flips a bit in the vernac state, so that:
  - if `cond` is false all sentences up to `#with.` are done normally, and `#with.` is classified as VtNow, typed as VtDefault. It's execution flips a bit in the vernac state, so that:
    - the parser accepts any sentence terminated with `. ` (as PG and CoqIDE do, see (*) later)
    - all sentences are classified/typed as no-op but for `#end.` which goes back to normal (VtNow, restores the normal parser)
  - if `cond` is true all sentences up to `#with.` are classified/typed as no-op; `#with.` flips a big so that all sentences following it are executed normally, `#end` is a no op.

As I wrote in the CEP, it is important that the text invoking the generator (between `#with` and `#end`) is not really parsed, since it may require parsing rules only available when the generator is available. Coq can as all accept `.*#end.` as a single sentence, but old UIs which do not delegate parsing to Coq are going to send sentences up to the first `.` and Coq has so "ignore" them (*).

In the simple sequential model things are pretty easy. For the STM & co, VtNow is enough to declare that `#with.` can change the parser (to essentially ignore all that follows). Note that the value of the conditionals is fixed, does not change as execution goes.

Hope it is clear, I can detail more if you like.

SkySkimmer:
`<=` isn't real attribute syntax, do you want to change that?

gares:
I think so, also `<` is not there IIRC, but I'm open to any other proposal.

maximedenes:
What's the process to deal with abandoned CEPs? Shouldn't we close the PR?

gares:
Ftr I did implement a simple version of this for Elpi vernacular commands and it was quite helpful.

PR #56, opened 2021-04-20
Visual debugging in secondary files

jfehrle:
Looking for general feedback--is this a reasonable plan?

Rendered: https://github.com/jfehrle/ceps/blob/debugger_gui/text/056-debugger-gui.md

### DIFF
--- a/text/056-debugger-gui.md
+++ b/text/056-debugger-gui.md
@@ -1,0 +1,103 @@
+- Title: Visual debugger: debugging in secondary script files
+
+- Drivers: Jim Fehrle
+
+----
+
+# Summary
+
+Proposes changes to CoqIDE's GUI layout and behavior to support stopping
+the debugger in secondary script files (i.e. files other than the topmost
+script), which may be requested by "step-in" requests or breakpoints in the
+secondary file.
+
+Discuses how to ensure that breakpoint locations shown in the GUI remain
+consistent with those in processed or compiled code as the user edits the
+source script(s).
+
+# Analysis
+
+## GUI
+
+The simple 1:1:1 relationship between script panels, Messages panels and
+coqidetop sessions/debuggers from [#13783](https://github.com/coq/coq/pull/13783)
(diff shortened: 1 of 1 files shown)


jfehrle:
Proposed appearance of the GUI with debugger panels:

debugger3

Appearance of breakpoints (red background) and the debugger stopping point (dark blue)--already implemented but not submitted:

image

text/056-debugger-gui.md
```diff
@@ -0,0 +1,103 @@
...
+
+Or consider the case in which two scripts both depend on `secondary.v`.
+After starting the first script, the user modifies and recompiles `secondary.v`
+and starts the second script.  To handle this case correctly, CoqIDE would need
+to maintain one mapping *per session*.
+
+On the other hand, we shouldn't go out of our way to support multiple
+simultaneous debug sessions--following a single debug session already requires
+very close attention by the user.
+
```
jfehrle, line 84: Another detail: If two debug sessions stop in the same file, how do you highlight 2 stopping points?  The stopping point highlight would need to change based on the which debugger tab is active.

Alizter:
As we discussed in the coq call, I think it would be better to only allow one debug session per process of coqide. This will simplify the design and avoid headache for the user.

jfehrle:
@Zimmi48 Perhaps you would merge this?

Zimmi48:
@jfehrle is this up-to-date with the current implementation plans? If not, maybe you want to add a note on what changed.

jfehrle:
Yes, it's current.  OTOH this CEP is meant to give some analysis, suggest alternative approaches and facilitate a discussion in order to get feedback on the ideas.  It's now a historical artifact.  It's not a spec, nor is it user or internal documentation, so updating it seems unnecessary.  Nor do I consider it as constraining the implementation, either--if I come up with better ideas during the development process, I may adopt them and then (if necessary) defend those changes during the review process.

The last part of the user documentation (#14381) reflects the discussion of the CEP.

PR #58, opened 2021-08-29
A few propositions about the module system

herbelin:
This CEP discusses a few limitations of the module system and give propositions to address them.

[Rendered here](https://github.com/herbelin/ceps/blob/module-limitations/text/modules.md)

### DIFF
--- a/text/modules.md
+++ b/text/modules.md
@@ -1,0 +1,78 @@
+- Title: Addressing a few limitations of the module system
+
+- Author: Hugo Herbelin
+
+----
+
+# Summary
+
+The module system of Coq has several roles:
+- name space management and visibility (`Module`, `Import`, `Export`, qualified names)
+- modular development of theories (modules parametric over module specifications, `:`, `<:`)
+- building of large theories by combination (`with`, `<+`, `Include`)
+- sealing of an implementation under an interface for the purpose of extraction towards the OCaml module system (`Module M:T:=S`)
+
+Though inherited from the OCaml module system, it differs from the latter by the absence of syntactic differences between interfaces and implementations. In Coq, types are terms, interfaces can have arbitrary kinds of definitions, and implementations can have axioms.
+
+This CEP tries to address a few limitations of the module system:
+1. identifying modules canonically defined by their implementation and modules canonically defined by their interface
+2. giving direct access to implementations sealed behind interfaces without going through extraction
+3. integrating ideas inherited from OCaml's module system
+4. providing more algebraic combinators to build theories
+
+# Propositions
+
+## Towards identifying modules canonically defined by their implementation and modules canonically defined by their interface
+
+Despite the absence of syntactic differences between interfaces and implementations, Coq maintains a somehow artificial distinction between a (so-called) "declared" module, that is a module characterized by its interface and whose implementation is canonically defined to be the same as the interface, and a "defined" module, that is a module characterized by its implementation and whose interface is canonically defined to be the same as the implementation. Here are examples showing the limitation:
+- In extraction:
+  ```coq
+  Require Import Extraction.
+
+  Module M. Definition a:=0. End M.
+  Extraction M. (* ok *)
+
+  Module Type T. Definition a:=0. End T.
+  Declare Module N : T.
+  Fail Extraction N. (* currently unsupported *)
+  ```
+- In module definitions:
+  ```coq
+  Module M. Axiom a:nat. End M.
+  Fail Module N := M with Definition a:=0. (* "with" not allowed for modules *)
+
+  Module Type T. Axiom a:nat. End T.
+  Fail Module N := T with Definition a:=0. (* T not a module *)
+
+  Declare Module N : T with Definition a:=0. (* ok *)
+  ```
+  even though all three declarations basically build the same module.
+
+  Note: the syntactic limitation of the first case is addressed in [#11897](https://github.com/coq/coq/pull/11897).
+
+In the implementation, these limitations come from the difference between `MExpr(params,expr,None)` (= a module with interface canonically identified with the implementation) and `MType(params,expr)` (= a module with implementation canonically identified with the interface). This difference goes to `module_body` by giving either the value `FullStruct` or the value `Abstract` to the field `mod_expr`.
+
+Both issues would be solved by identifying the two constructions (i.e. rebinding `MType(params,expr)` to `MExpr(params,expr,None)`) and identifying `FullStruct` and `Abstract`.
+
+## Giving direct access to implementations sealed behind interfaces without going through extraction
+
+In Coq, only the interface of a module is relevant for typing a development dependent on a module. In particular, sealing implementations behind interfaces definitely prevents being able to access the contents of an implementation. The latter is then be seen only by extraction which can export the implementation to a programming language, such as OCaml, Haskell, Scheme, etc.
+
+But the implementation is still there!
+
+The proposal is to provide a command `Magic Eval reduction in term` which evaluates `term` by referring to the implementation of the modules mentioned in `term` rather than to their interface. For instance, in:
+```coq
+Module Type T. Axiom f:nat->nat. End T.
+Module N:T. Definition f:=Nat.pred. End N.
+```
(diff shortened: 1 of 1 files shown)


text/modules.md
```diff
@@ -0,0 +1,78 @@
...
+```coq
+Module Type T. Axiom f:nat->nat. End T.
+Module N:T. Definition f:=Nat.pred. End N.
+```
+calling `Magic Eval compute in N.f 1.` would produce `0`.
+
+## Ideas inherited from OCaml's module system
+
+OCaml has two module system features which Coq does not have:
+- Private inductive types
```
Alizter, line 73: Don't we have private inductive types in coq?

text/modules.md
```diff
@@ -0,0 +1,78 @@
...
+- Axiomatic variant of `Module Type` providing a polymorphically-typed module system
+
+## Providing more algebraic combinators to build theories
```
Alizter, line 76: Have you got any more combinators in mind?

Alizter (approved):

text/modules.md
ppedrot: Yes, but it does not mean the same and it's not a kernel feature (read: it's a broken hack).

silene:
In Why3, we do not distinguish between module implementations and module interfaces. For instance, your following example works fine once translated to WhyML:
```coq
Module M. Axiom a:nat. End M.
Module N := M with Definition a:=0.
```

```whyml
module M
  constant a : int
end

module N
  constant a : int = 0
  clone M with constant a
end
```

The system works rather well, including private types and extraction. So, perhaps, some ideas could be reused for Coq. That said, Why3 has a much simpler type system, so hard to tell how bad it would behave if Coq's conversion was thrown into the mix.

Some details are given in this paper: https://hal.inria.fr/hal-02696246

herbelin:
The simplicity of the module system of Why3 is appealing (and somehow "to the point"). If I understand correctly, it can emulate higher-order functors, `<:` and even `:` with some discipline (e.g. to extract `M:T` I guess you have to explicitly tell that you want to extract `M` coupled with `T` since there is no such primitive concept of module sealed behind an interface). It has no `Module Type` but one can use `clone` to treat a module as a module type (i.e. so that it is generative). Is that a correct understanding? (As a side note, the frame type theory of X. Montillet, A. Mahboubi and C. Cohen has similar ideas.)

text/modules.md
herbelin: Not really. I was mostly thinking at the oppportunity to integrate `Include` and the quite powerful `<+` in the kernel.

herbelin:
By the way, shouldn't your example (morally) use Why3's `use` rather than `clone` so that (if there were some) the inductive declarations in `M` are aliased in `N`, as it would be in Coq.

silene:
> Is that a correct understanding?

Yes.

As for `M:T`, this is getting added to Why3 and will presumably be part of the next release. The meaning is intermediate between `:` and `<:` in Coq, that is, `M` is sealed when doing proofs (only its interface `T` is available), but its implementation is actually available for interpretation and extraction. I am not sure whether this would translate well to Coq, since types and proofs are plain values there.

> shouldn't your example (morally) use Why3's `use` rather than `clone`

No. As in Coq and OCaml, Why3's modules act both for namespacing and as a higher-order type system. Directive `use` is for the former, while `clone` is for the latter. You would only use `use` for a fully concrete module implementation (or one with abstract symbols that are not meant to be instantiated, e.g., `int.Int` or `real.Real`), so kind of like `Require` in Coq.

But you are right that it causes an issue if you want inductive types to be shared (since module cloning is generative). In that case, you would use the same trick as in Coq (which makes it just as tedious):

```whyml
module K
  type foo = Foo
end

module M
  use export K
  constant a : int
end
```

Now, any clone of `M` will contain a symbol `foo` that is synonymous with `K.foo` (and same for `Foo`).

text/modules.md
SkySkimmer: ocaml `private` is used to enforce guarantees about the values in the type.
In Coq this is can done by explicitly manipulating proofs about the values in the type, so there seems to be no need to for ocaml-style `private`.
Also I'm not sure how it would work, consider 
~~~coq 
Module Import SortedList.

Private(ocaml) Inductive t := Nil | Cons : nat -> t -> t.

Definition nil := Nil.
Definition cons x l := (* add x to l keeping it sorted *).
End SortedList.
Compute nil. (* Nil : t ??? *)
~~~

text/modules.md
```diff
@@ -0,0 +1,78 @@
...
+Module Type T. Axiom f:nat->nat. End T.
+Module N:T. Definition f:=Nat.pred. End N.
+```
+calling `Magic Eval compute in N.f 1.` would produce `0`.
+
+## Ideas inherited from OCaml's module system
+
+OCaml has two module system features which Coq does not have:
+- Private inductive types
+- Axiomatic variant of `Module Type` providing a polymorphically-typed module system
```
SkySkimmer, line 74: Can this lead to a type-in-type inconsistency?

text/modules.md
silene: I suppose opaque types would have abstract values, as in OCaml: `<abstract> : t`.

text/modules.md
```diff
@@ -0,0 +1,78 @@
...
+
+  Declare Module N : T with Definition a:=0. (* ok *)
+  ```
+  even though all three declarations basically build the same module.
+
+  Note: the syntactic limitation of the first case is addressed in [#11897](https://github.com/coq/coq/pull/11897).
+
+In the implementation, these limitations come from the difference between `MExpr(params,expr,None)` (= a module with interface canonically identified with the implementation) and `MType(params,expr)` (= a module with implementation canonically identified with the interface). This difference goes to `module_body` by giving either the value `FullStruct` or the value `Abstract` to the field `mod_expr`.
+
+Both issues would be solved by identifying the two constructions (i.e. rebinding `MType(params,expr)` to `MExpr(params,expr,None)`) and identifying `FullStruct` and `Abstract`.
```
Blaisorblade, line 55: @herbelin Are you proposing mixin modules, a bit like in Scala or MixML? I’ve often wanted them, _and_ I’ve actually spent a couple of years studying the Scala case.

Here’s a small example:

```coq
Module Type Base.
  Axiom a : nat.
  Definition b := 1 + a.
End Base.

Module Derived1 <: Base.
  Definition a : nat := 0.
  Include Base.
  (* Include should _merge_ the axiom `a` with the `Definition `` as long as the definition is a subtype of the interface — if Coq satisfies the usual metatheory of subtyping, definitions using `a` should still typecheck and need not be rechecked, except for universe constraints since those are not modular. *)
End Derived1.

Module Derived2 <: Base.
  Include Base.
  Fail Definition a : nat := b. (* This creates a new kind of infinite loop, which must be forbidden. *)
  Definition a : nat := 0.
  (* Such things work in Scala but is harder, since we have to _replace_ a definition; rejecting this might be fine here, but not so well with FancyBase below. To make this work, we could replace [Include] by some other “inheritance” mechanism. *)
End Derived2.

Module FancyBase.
  Include Base.
  Axiom c : Fin b.
  Definition d := (c, c).
End FancyBase.

Module FancyDerived1 <: FancyBase.
  Definition a : nat := 0.
  Definition c : Fin 1 := Fin.FZ. (* We cannot write [Fin b], and we might be unable to write this definition at all. *)
  Include FancyBase.
End FancyDerived1.

Module FancyDerived2 <: Base.
  Include FancyBase.
  (* Now we can fill in the axioms *)
  Definition a : nat := 0.
  Definition c : Fin b := Fin.FZ.
End FancyDerived2. (* *)
```

This code can be translated to avoid mixing modules — Coq’s numerical hierarchy shows how. But the encoding is pretty annoying, since you can never mix axioms and definitions if the axioms are to be later instantiated.

I also realize that working out the details is not trivial; but I am not sure what you get if you stop short of mixin modules.

text/modules.md
Blaisorblade: Could this proposal be clarified? Is this `Declare Module Type Foo.` where `Foo` is then an abstract signature?

PR #59, opened 2021-08-31
Unification structure

herbelin:
This CEP is about organizing knowledge about the (evarconv) unification algorithm of Coq in the hope to draw directions for the future.

It is a rather extensive first draft but at the same time a rather naive one. A lot should be done to make it even more comprehensive, collecting experiences and contributions of everyone.

[Rendered here](https://github.com/herbelin/ceps/blob/unification-structure/text/unification.md)

### DIFF
--- a/text/unification.md
+++ b/text/unification.md
@@ -1,0 +1,268 @@
+- Title: The structure of the unification algorithm
+
+----
+
+# Summary
+
+Unification is a central component of a proof assistant, used for type inference, inference of implicit information, tactic application.
+
+The unification algorithms in Coq have grown organically by need. The code is currently rather convoluted and somehow fragile, with small changes in some parts of the code having easily side effects on other parts of the code.
+
+Sozeau and Ziliani 2015 and 2016 designed a unification algorithm based on a clear structure. This CEP is about describing the components of the current unification algorithm in the direction of re-articulating it around a similarly well-understood structure like Sozeau and Ziliani did.
+
+A priori the following basic architecture is intended to be preserved:
+- a part canonically splitting complex unification problems into atomic ones (e.g. `(0,?f 0) == (?m,(0,0))` into non-ambiguous assignment `?m := 0` and multi-solution `?f 0 == (0,0)`) with linear complexity
+- a part applying heuristics when several solutions are possible (e.g. resolving `?f := fun x => (x,x)` rather than `?f := fun x => (0,0)` or `?f := fun x => (0,x)` or `?f := fun x => (x,0)`), possibly using backtracking
+
+# Basic background on unification
+
+## Existential variables
+
+Unification relies on *existential variables* which are defined in a context which is specific to each of them. The declaration of an existential variable takes the form:
+```
+a₁:A₁,...,aₙ:Aₙ ⊢ ?e : B
+```
+where:
+- `a₁:A₁,...,aₙ:Aₙ` is the defining context of the existential variable (called `evar_hyps` in the implementation, it is made of named declarations, possibly with local definitions too, so, shortly a `named_context`)
+- `B` is its type (`evar_concl`)
+
+To be used in a term, existential variables need to be applied to a substitution of its context. So, they take the form `?e@{a₁:=t₁,...,aₙ:tₙ}` where `a₁:=t₁,...,aₙ:=tₙ` is a *substitution* (sometimes called *instance*, often written `args` in the code).
+
+## Existential variables map
+
+A collection of declarations of existential variables is conventionally written `Σ` (an `evar_map` in the code, using either `sigma` or `evd` to refers to it).
+
+## Unification problems
+
+Given an existential variable map, *unification problems* are equations of the form `Γ ⊢ t₁ =?= t₂` or inequations of the form `Γ ⊢ t₁ ≤? t₂` where `t₁` and `t₂` are terms possibly mentioning existential variables with their substitution and both terms living in a context of variables `Γ` (that is the juxtaposition of a `named_context` and of a `rel_context`). An inequation corresponds to a subtyping problem between types (e.g. `Prop ≤ Type`). Note that we assume the existence of a global context of inductive types, definitions and axioms common to the definitions of existential variables and to the unification problems.
+
+A unification algorithm turns a `Σ` and a collection of unification problems into either an impossibility of satisfaction of the equations, or in a collection of (well-typed) definitions for the existential variables such that all problems are satisfied, or a residual set of unification problems that the algorithm is neither able to invalidate nor to satisfy (since unification in general is undecidable in the Calculus of Inductive Constructions).
+
+## Reflexive unification problems
+
+We call *reflexive unification problem* an equation of the form `Γ ⊢ ?e@{subst} =?= ?e@{subst'}`.
+
+## Atomic unification problems
+
+We call *atomic unification problem* an equation of the form `Γ ⊢ ?e@{subst} =?= ?e'@{subst'}` or an inequation of the form `Γ ⊢ ?e@{subst} ≤? ?e'@{subst'}`.
+
+## Simple unification problems
+
+We call *simple unification problem* an equation of the form `Γ ⊢ ?e@{subst} =?= t` or an inequation of the form `Γ ⊢ ?e@{subst} ≤? t`. Unification basically reduces arbitrary complex unification problems to simple ones which in turn may generate new arbitrary complex problems.
+
+## Simplification of reflexive unification problems
+
+A reflexive unification problem induces constraints on the possible definitions of `?e`: once `?e` is defined, it may generate new unification problems between terms of the instances that contain existential variables.
+
+A reflexive unification problem can be simplified as follows:
+- Variables of the context whose instances in the substitution are provably incompatible (e.g. because corresponding to distinct head normal forms) can be removed from the context.
+- If both instances are free of existential variables, the problem can be dropped since its resolution will eventually only create equations between convertible existential-variables-free terms.
+
+## Simplification of atomic unification problems
+
+An atomic unification problem can be simplified as follows:
+- Variables of the context of the existential variables that are bound to terms mentioning variables of the context of the equations in non-erasable position in only one of the two instances can be restricted since there would be no way to make these terms match a subterm of the instance of the other existential variables; otherwise said, both existential variables can be restricted on the (well-typed closure of the) subsets of their context which intersect.
+- If the problem is an inequation, it can only be postponed since one variable could be instantiated e.g. with `Prop` and the other with `Type` without them to be equal.
+- If one of the instances mention only distinct variables on the subset where there is an intersection, *and* the problem is not an inequation, the corresponding existential variable can be defined with the other one without loss of solutions
+
+## Basic algorithm for resolving simple unification problems
+
+Solving non atomic simple unification problems relies on basically two concepts (see Huet 1975, `MATCH` algorithm):
+- *Projection*, that is, solving `Γ ⊢ ?e@{a:=t,b:=t} =?= t` by either setting `Δ ⊢ ?e := a` or `Δ ⊢ ?e := b` so that `?e@{a:=t,b:=t}` indeed evaluates in `Γ` to `t`.
+
+  In the more general case, a projection requires to solve general unification problems, that is, to know if `Γ ⊢ ?e@{a:=t} =?= u` can be solved by projecting `a` requires to find a context `E` such that `Γ ⊢ E[t] =?= u` is resolvable (up to reduction of `E[t]`). This can be done in a systematic way in algebraic-type-free λ-calculus by reasoning on the normal forms of `t` and `u` and using type constraints. It is not clear however that a systematic study of the process and of the form of evaluation ontexts has already been done somewhere in the presence of algebraic types.
+
+  For instance, in the presence of algebraic types, there is a specific incarnation of projection steps involving constructors. For instance, the problem `Γ ⊢ match ?e with S n => t | O => u end =?= v` can be solved (among other resolution strategies) by defining `?e := S ?n` for a new existential variable `?n` and solving `Γ ⊢ t[n:=?n] =?= v` or defining `?e := 0` and solving `Γ ⊢ u =?= v`.
+
+  We call projection of variables the subcase when `u` is a variable and `E` is empty. In practice, Coq implements only projections of variables and projections of terms of the form `C t₁ ... tₙ` for `C` a constructor, with `E` empty (but PR [#14798](https://github.com/coq/coq/pull/14798) extends this to `E` made of nesting of `match` and projections).
+
+- *Imitation*, that is, imitating the structure of `t`, or part of it, and generating subproblems about the subterms of the imitated part of `t`. 
+
+  For instance, if `C` is a constructor, the equation `Γ ⊢ ?e@{subst} =?= C t₁ ... tₙ` reduces to the collection of equations `Γ ⊢ ?eᵢ@{subst} =?= tᵢ` by declaring new existential variables `?eᵢ` of appropriate type in the same context as `?e` and by setting `?e := C ?e₁@{dom(Δ)} ... ?eₙ@{dom(Δ)}`. 
+
+  Similarly, if the head symbol of `t` is a binder, new existential variables are generated for the subterms of the binder in a context extended with the new binder.
+
+## Efficiency non-determinism
+
+The non-determinism comes from the potentially different ways to project and from the ability to also sometimes imitate. For instance, there are 2 possible projections and one possible imitation to resolve `Γ ⊢ ?e@{a:=t,b:=t} =?= t` for `t` variable-free.
+
+There is also a form of non-determinism which comes from reducing unification problems or not in normal form:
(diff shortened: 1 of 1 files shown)


text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+## Simple unification problems
+
+We call *simple unification problem* an equation of the form `Γ ⊢ ?e@{subst} =?= t` or an inequation of the form `Γ ⊢ ?e@{subst} ≤? t`. Unification basically reduces arbitrary complex unification problems to simple ones which in turn may generate new arbitrary complex problems.
+
+## Simplification of reflexive unification problems
+
+A reflexive unification problem induces constraints on the possible definitions of `?e`: once `?e` is defined, it may generate new unification problems between terms of the instances that contain existential variables.
+
+A reflexive unification problem can be simplified as follows:
+- Variables of the context whose instances in the substitution are provably incompatible (e.g. because corresponding to distinct head normal forms) can be removed from the context.
```
JasonGross, line 58: This optimization seems unsound?
Consider `x : nat |- ?e : nat` and the pair of unification problems `|- ?e@{x:=O} == ?e@{x:=S O}` and `x : nat |- ?e@{x:=x} == match x with O => O | S _ => O end`.  The only valid solution of this is `?e := match x with O => O | S _ => O end`, which refers to `x`.
And in fact Coq fails to solve such a problem:
```coq
Check let e := (fun x => ?[e]) in
      eq_refl : (e O, e) = (e (S O), fun x => match x with O => O | S _ => O end).
(* Error:
In environment
e := fun _ : nat => ?e : nat -> nat
The term "eq_refl" has type "(e 0, e) = (e 0, e)" while it is expected to have type
 "(e 0, e) = (e 1, fun x : nat => match x with
                                  | 0 | _ => 0
                                  end)" (cannot instantiate "?e" because "x" is not in its scope).
*)
```

JasonGross: Reported as https://github.com/coq/coq/issues/15110

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+- If both instances are free of existential variables, the problem can be dropped since its resolution will eventually only create equations between convertible existential-variables-free terms.
```
JasonGross, line 59: I'm not sure I understand this one.  Is conversion run again after unification completes?  How do you get convertibility constraints?  Also, this seems overly optimistic, consider `x : nat |- ?e : nat` and `|- ?f : nat` and the pair of unification problems `|- ?e@{x:=O} == ?e@{x:=S O}` and `x : nat |- ?e@{x:=x} == match x with O => O | S _ => ?f end`.  We must instantiate `?e` with `match x with O => O | S _ => ?f end`, and only after this substitution do we get `|- O == ?f` out of the first unification problem, which it seems would have been dropped according to this rule.

It seems that Coq can solve this problem, though?
```coq
Check let e := (fun x => ?[e]) in
      let f := ?[f] in
      eq_refl : (e, e O) = (fun x => match x with O => O | S _ => f end, e (S O)).
```

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Simplification of atomic unification problems
+
+An atomic unification problem can be simplified as follows:
+- Variables of the context of the existential variables that are bound to terms mentioning variables of the context of the equations in non-erasable position in only one of the two instances can be restricted since there would be no way to make these terms match a subterm of the instance of the other existential variables; otherwise said, both existential variables can be restricted on the (well-typed closure of the) subsets of their context which intersect.
```
JasonGross, line 64: What does "non-erasable position" mean?  What does "restricted" mean?

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+A. There is a function `Evarsolve.solve_simple_eqn` to solve simple unification problems but:
+1. It may apply heuristics too early:
+   1. when trying to solve `?e@{subst} := t`, it ignores the terms of the substitution which are neither a variable nor an applied constructor; e.g. when trying to solve `?e@{a:=Prop} := Prop`, it assumes that imitation is the only possibility but that's wrong
+
+   On the other side, projecting terms which are not in head normal form may be arbitrarily costly:
+   - so, we may consider projecting ...
+   - as a heuristic, we may still rely on syntactic equality so as, e.g., to use a projection in `?e@{ack 4 4} := ack 4 4`
+
+2. It may fail to find possible solutions:
+   1. there arye valid circular instances that it is not able to accept (see ongoing PR [#14746](https://github.com/coq/coq/pull/14746))
```
JasonGross, line 202:
```suggestion
   1. there are valid circular instances that it is not able to accept (see ongoing PR [#14746](https://github.com/coq/coq/pull/14746))
```

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+
+**Extensions**
+- Extend canonical structures to canonical equipments based on knowing two or more projections?
+- Reorganize canonical structures so that it more generally solves equations of the form `proj₁ᵢ₁ (... projₙᵢₙ ?e ...) = u`. On one side, it should a priori give more discriminating solutions. On the other side, it would allow to treat canonical structures only in the flexible case.
+- Extend first-order unification to `match` term, e.g. to resolve `if true then t else u =?= if ?e₀ then ?e₁ else ?e₂` by setting `?e₀ := true`, `?e₁ := t` and `?e₂ := u` (see wish [#12780](https://github.com/coq/coq/issues/12780)).
+- Existential variables for sorts can sometimes not be instantiated by `Prop` even in situation where it would be legitimate to do so (see PR [#7369](https://github.com/coq/coq/pull/7369)).
+
+**Optimizations**
+- Systematically eta-reduce the solutions to existential variables (with all kinds of eta). On one side, this seems to be the natural expectation. On the other side, this would avoid caring about whether it might be dommageable to eta-expand this or that solution in the unification algorithm.
```
JasonGross, line 258:
```suggestion
- Systematically eta-reduce the solutions to existential variables (with all kinds of eta). On one side, this seems to be the natural expectation. On the other side, this would avoid caring about whether it might be harmful to eta-expand this or that solution in the unification algorithm.
```
Translating French to English with Google

text/unification.md
JasonGross: If I'm understanding correctly, this again seems incorrect for the same reason as above.  We might have consider `x : nat |- ?e : bool` and `x : nat |- ?e' : bool` and the triple of unification problems `x, y : nat |- ?e@{x:=S x} == ?e'@{x:=S y}` and `x : nat |- ?e@{x:=x} == match x with O => true | S _ => false end` and `x : nat |- ?e@{x:=x} == match x with O => true | S _ => false end`. We must instantiate `?e` and `e'` both with `match x with O => true | S _ => false end`.
```coq
Check let e := (fun x => ?[e]) in
      let e' := (fun x => ?[e']) in
      fun x y => eq_refl
                 : (e (S x), e, e') =
                   (e' (S y), (fun x => match x with O => true | S _ => false end),
                    (fun x => match x with O => true | S _ => false end)).
(* Error:
In environment
e := fun x : nat => match x with
                    | 0 => true
                    | S _ => false
                    end : nat -> bool
e' := fun _ : nat => false : nat -> bool
x : nat
y : nat
The term "eq_refl" has type "(e (S x), e, e') = (e (S x), e, e')" while it is expected to have type
 "(e (S x), e, e') =
  (e' (S y), fun x : nat => match x with
                            | 0 => true
                            | S _ => false
                            end, fun x : nat => match x with
                                                | 0 => true
                                                | S _ => false
                                                end)" (cannot unify "e'" and
"fun x : nat => match x with
                | 0 => true
                | S _ => false
                end").
*)
```

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+- If one of the instances mention only distinct variables on the subset where there is an intersection, *and* the problem is not an inequation, the corresponding existential variable can be defined with the other one without loss of solutions
+
+## Basic algorithm for resolving simple unification problems
+
+Solving non atomic simple unification problems relies on basically two concepts (see Huet 1975, `MATCH` algorithm):
+- *Projection*, that is, solving `Γ ⊢ ?e@{a:=t,b:=t} =?= t` by either setting `Δ ⊢ ?e := a` or `Δ ⊢ ?e := b` so that `?e@{a:=t,b:=t}` indeed evaluates in `Γ` to `t`.
+
+  In the more general case, a projection requires to solve general unification problems, that is, to know if `Γ ⊢ ?e@{a:=t} =?= u` can be solved by projecting `a` requires to find a context `E` such that `Γ ⊢ E[t] =?= u` is resolvable (up to reduction of `E[t]`). This can be done in a systematic way in algebraic-type-free λ-calculus by reasoning on the normal forms of `t` and `u` and using type constraints. It is not clear however that a systematic study of the process and of the form of evaluation ontexts has already been done somewhere in the presence of algebraic types.
```
JasonGross, line 73:
```suggestion
  In the more general case, a projection requires to solve general unification problems, that is, to know if `Γ ⊢ ?e@{a:=t} =?= u` can be solved by projecting `a` requires to find a context `E` such that `Γ ⊢ E[t] =?= u` is resolvable (up to reduction of `E[t]`). This can be done in a systematic way in algebraic-type-free λ-calculus by reasoning on the normal forms of `t` and `u` and using type constraints. It is not clear however that a systematic study of the process and of the form of evaluation contexts has already been done somewhere in the presence of algebraic types.
```

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+There is also a form of non-determinism which comes from reducing unification problems or not in normal form:
+- When resolving `Γ ⊢ c t =?= c' t'` with `c` and `c'` distinct unfoldable constants, there is a choice in reducing `c` or `c'` in the hope that the problem is eventually found satisfiable. This choice can lead to arbitrarily long computations so, controlling unfolding is important (there are similar problems when efficiently comparing the convertibility of two terms).
+- It is also useful to limit unfolding of constants so as to obtain solutions as close as possible to the original form of the terms involved.
+
+## Higher order pattern unification
+
+*Higher order pattern unification* (see Miller 1991 and Pfenning 1991) is a special case of simple unification problem when the substitution is made only of distinct variables. In this case, there is a single solution obtained by projecting the unique possible variable or imitating if not a variable.
+
+Pattern unification has various refinements (see e.g. Libal and Miller 2016) such as:
+- expecting distinct variables only among the variables that actually occur in the term to match
```
JasonGross, line 98: Note that this can cause the pattern solution to be incorrect if there are other variables in the context which are assigned to terms which are headed by a constructor, because we may choose whether or not to match on them.  (Consider `x, y : bool |- ?e` and `x : bool |- ?e@{x:=x,y:=true} == x`, we can solve this with either `x` or with `if y then x else _`, and the latter may be manded by a future unification problem.)

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+- identifying arguments that are irrefutable patterns over variables at their leaves (e.g. `(x,(y,z))` with those variables (i.e. a form of curryfication) (see PR [#14798](https://github.com/coq/coq/pull/14798))
```
JasonGross, line 99: This is only guaranteed to generate a good solution either (a) in the presence of eta conversion for the record type, or (b) when no subterm of the constructor tree, other than the leaves, occurs in the RHS.  e.g., `x, y, z |- ?e@{x:=(x,(y,z))} == (y,z)` can be solved with either `snd x` or `(fst (snd x), snd (snd x))`.

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+
+  It is actually not fully clear to me what are all the ins and outs of first-order unification.
+  - In situations like `Q ?x =?= Q t` with `?x` erasable when reducing `Q ?x`, first-order unification gives a hint on how to infer `?x` and this seems to be a useful heuristics.
+  - In situations like `?P t =?= Q t`, it tells not to care about possible instances of `t` in `Q` and this is more disputable (e.g., it is a priori unexpected that `?P 0 =?= (0=0)` is solved into `?P := fun x => 0=x` rather than into `?P := fun x => x=x`).
+
+- Second-order unification heuristics: e.g. to resolve `?P t =?= u` by selecting a number of occurrences of subterms unifiable to `t` in `u` (see `Evarconv.second_order_matching`). Such a heuristic is central to any form of reasoning by induction. As far as I can recap, it is currently used in Coq only for tactics.
+  - One issue is to recognize subterms: we can assume `t` free of existential variables and reason modulo convertibility, or we can reason using syntactic equality, or syntactic equality discarding non-essential, i.e. reinferable arguments (i.e., morally the implicit arguments).
```
JasonGross, line 115: I think this should be subsumed by introducing constructors of types with eta.  That is, if we have `?P t =?= u`, we can always instantiate `?P` with `fun x => ?Q`, turning the application into a substitution.  And if we have `fst ?x =?= u` (or `match ?x with pair x y => Q end =?= u`) and primitive projections / eta are turned on, it should always be safe to instantiate `?x` with `(?a, ?b)` (though we probably want to run canonical structures first, and we maybe want to make sure that `u` doesn't reduce to something that structurally matches such a projection/match, because otherwise we may forget information that we want to remember for first-order heuristics)

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+  - Another issue is that typing constraints induce equivalence classes of subterms that have to be abstracted altogether.
```
JasonGross, line 116: What does this mean?  Or what's an example?

text/unification.md
JasonGross: Oh, I see this is mentioned below

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+- Existential variables for sorts can sometimes not be instantiated by `Prop` even in situation where it would be legitimate to do so (see PR [#7369](https://github.com/coq/coq/pull/7369)).
+
+**Optimizations**
+- Systematically eta-reduce the solutions to existential variables (with all kinds of eta). On one side, this seems to be the natural expectation. On the other side, this would avoid caring about whether it might be dommageable to eta-expand this or that solution in the unification algorithm.
+- Give an order between constants which would allow to predict which side to reduce in `E[c] =?= E'[c]`.
+- Keep constants unfolding to fixpoints folded so as to concentrate on the essential and reduce the size of expressions, including in evaluation contexts (i.e. stacks).
+- Do more aggressive restrictions of contexts (e.g. Sozeau and Ziliani seem to take the downwards well-typed closure of restrictions while Coq takes the upwards well-typed closure - knowing that taking the closure by dependency requires in theory to take care of erasable occurrences)?
```
JasonGross, line 261: What's the difference between downwards and upwards here?  (Or what's an example?)

text/unification.md
herbelin: It seems I meant "If both instances are pointwise existential-variable-free compatible terms". So, with your previous remark, it should be "If both instances are pointwise existential-variable-free compatible neutral terms".

Your example works because `e := fun x => match x with O => O` is solved first. If you invert the order of unification problems:
```
Check let e := (fun x => ?[e]) in
      let f := ?[f] in
      eq_refl : (e O, e) = (e (S O), fun x => match x with O => O | S _ => f end).
```
it fails.

text/unification.md
herbelin: This should be formulated clearly. It is `among the variables`, so the presence of `true` makes that it is not a pattern (see the `Option.List.map` in `Evarsolve.is_unification_pattern_evar`; the restriction to actual dependencies in `get_actual_deps` comes after).

text/unification.md
herbelin: Right.

text/unification.md
herbelin: By `downwards` I mean the largest well-formed context which exclude the forbidden variables. By `upwards`, I mean the smallest well-formed context which include the allowed variables.

herbelin (commented):
Thanks a lot for the review! I'll answer the remaining questions when I'll have more time.

text/unification.md
JasonGross: Hmmm, it seems that upwards is more powerful unless we have reduced enough that no further occurrences of forbidden variables can be erased by reduction.  Consider `T : Type, f : T -> Type, x : T |- ?e : Type`, and the pair of problems `A, B : Type, f : bool -> Type, x : bool |- ?e@{T:=(fun _ => bool) A,f:=f,x:=x} == ?e@{T:=(fun _ => bool) B,f:=f,x:=x}` and `f : bool -> Type, x : bool |- ?e@{T:=bool,f:=f,x:=x} == f x`.  IIUC, upwards closure would keep `T`, while downwards would remove it?  (Do you have an example not involving reduction where they differ where upwards closure does better than downwards closure?)

JasonGross: But maybe this is what you meant by "erasable"?

text/unification.md
```diff
@@ -0,0 +1,268 @@
...
+- Existential variables for sorts can sometimes not be instantiated by `Prop` even in situation where it would be legitimate to do so (see PR [#7369](https://github.com/coq/coq/pull/7369)).
+
+**Optimizations**
+- Systematically eta-reduce the solutions to existential variables (with all kinds of eta). On one side, this seems to be the natural expectation. On the other side, this would avoid caring about whether it might be harmful to eta-expand this or that solution in the unification algorithm.
+- Give an order between constants which would allow to predict which side to reduce in `E[c] =?= E'[c]`.
+- Keep constants unfolding to fixpoints folded so as to concentrate on the essential and reduce the size of expressions, including in evaluation contexts (i.e. stacks).
+- Do more aggressive restrictions of contexts (e.g. Sozeau and Ziliani seem to take the downwards well-typed closure of restrictions while Coq takes the upwards well-typed closure - knowing that taking the closure by dependency requires in theory to take care of erasable occurrences)?
+- Filters need to be better understood: why not to directly restrict the context more often?
```
JasonGross, line 262: Is this a performance optimization?

PR #65, opened 2022-09-12
[CEP] Proposal for libobject API refactoring

ejgallego:
Rendered version [here](https://github.com/ejgallego/ceps/blob/libobject/text/libobject-refactor.md)

### DIFF
--- a/text/libobject-refactor.md
+++ b/text/libobject-refactor.md
@@ -1,0 +1,237 @@
+- Title: Libobject metadata and state redesign
+
+- Driver: Emilio J. Gallego Arias
+
+- Status: Draft, (partial) PR available, PR tested
+
+---
+
+# Summary and Objectives
+
+Coq's module system is (mainly) implemented by an abstract notion of
+"object" (`Libobject.t`), which corresponds to an existential type `t`
+which has to be (implicitly) `Marshall`able, plus a set of operations
+`Libobject.object_declaration` which specify the effects the object
+will perform when affected by several operations such as `Require`,
+`Import`, etc... in a fully imperative (that is to say `unit -> unit`)
+way.
+
+Examples of such objects are declarations, implicit arguments, modules
+themselves, implicit arguments and quantifiers, classes, notations, ...
+
+There are quite a few downsides to this approach, in particular:
+
+- there is no mechanism for objects to declare common meta-data such
+  as documentation, origin (user or auto-generated), location, etc...
+
+- there is no mechanism for objects to declare common behavior such as
+  registering with the nametab, declaring tables or maps, etc..
+
+- there is no control of interaction of objects with the parser, kernel
+
+- there is no possibility to systematically explore the list of
+  objects currently in scope (other than hoping the ad-hoc internal
+  structures are somehow exposed)
+
+- in some cases (such as printing), we need to perform a costly (and
+  messy) global imperative state update as to simulate that a module
+  has been opened, due to the imperative nature of the module system
+  implementation
+
+These shortcoming are handled today by introducing ad-hoc effects in
+the imperative handler, but clients such as UI interfaces or Search
+have a hard time working in a clean way, the API state is far from
+satisfactory.
+
+This CEP proposes a remodelling of the `Libobject` API based on our
+experience with alternative Coq toplevels and interaction systems. In
+particular, we aim to:
+
+- split concerns in the module system implementation: serialization,
+  scope handling, object life-time and effects
+
+- provide a common API for indexing a retrieving object metadata
+
+- provide a common API for declaration of side-effects, such as
+  creating tables, references, registering with the kernel, the
+  parser, or the nametab
+
+- set the path for a functional implementation of some basic
+  operations of the module system, pushing the imperative state to the
+  upper layers
+
+- replace `.glob` files by a query to the object database, provide
+  rich documentation, jump to definition features
+
+## Coq DB Workshop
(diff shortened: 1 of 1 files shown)


text/libobject-refactor.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Loading
+
+`Declaremods` already stores the objects in a map indexed by their
+prefix, we propose to:
+
+- create a reverse map or maps to allow clients to query the `Data.t`
+  efficiently. In particular we could have reverse maps for
+  categories, names, locations. For now we implement a reverse map for
+  names, but more options could be desirable after more experimentation.
```
SkySkimmer, line 148: Does this refer to a branch somewhere or is more like "for now we propose to start with a reverse map for names"?

text/libobject-refactor.md
```diff
@@ -0,0 +1,197 @@
...
+`declare_object` API in favor a set of `Behavior.t` traits that an
+object can implement, in a functional way.
+
+The main goal here is to allow the upper layer to control the state,
+and have more flexibility in updating the state. This has many
+applications.
+
+In particular, we can distinguish a few common behaviors:
+
+- `Kernel` : objects with this trait provide a function to register with a `safe_env`
```
SkySkimmer, line 175: Note that currently I believe the only part of libobject which modifies the safe_env is the Require replay after closing section/module (in_require in library.ml).

text/libobject-refactor.md
```diff
@@ -0,0 +1,197 @@
...
+- `Named` : objects with this trait can register with the `Nametab`, in a uniform way
```
SkySkimmer, line 176: My understanding is that currently nametab itself is not uniform (there are separate tables for globref, modules, ltacs, some other stuff) so this may require a nametab redesign first.

ppedrot: FTR I think that unifying the various nametabs is gong the wrong way. It's a feature that each plugin can define its own set of names, so the design critically needs to account for this.

ejgallego: Yes, the idea is that when an object declares a nametab trait, it does so with their own nametab. Refactoring is indeed needed, you can see current progress here, but need to advance more https://github.com/ejgallego/coq/blob/f059e5a57e95c5c80a3c9026e91ba91936d90db5/library/nametab.mli

Ideally, if we think this is a good idea we would do such a change in two steps:
- we would first define the trait, which takes a module as a parameter and performs the imperative state
- we would then push the trait to be functional, and have `Declaremods` own the nametab.

text/libobject-refactor.md
ejgallego: Yup, that's correct. Tho the trait is meant to capture objects that call the diverse `Global.*` functions, so at some point the env can be owned by declaremods.

text/libobject-refactor.md
ejgallego: This is implemented in the prototype branch, cleaning it up to submit as we speak.

ejgallego: PR done, you can use (hack hack) `Print Graph` to print the reverse map

text/libobject-refactor.md
ejgallego: I've clarified this a bit in an update.

text/libobject-refactor.md
ejgallego: Clarified a bit.

gares:
Can you explain how the proposed approach solves/improves-on the downsides of the current approach, in particular:

> - there is no possibility to systematically explore the list of objects currently in scope (other than hoping the ad-hoc internal   structures are somehow exposed)

> - in some cases (such as printing), we need to perform a costly (and messy) global imperative state update as to simulate that a module has been opened, due to the imperative nature of the module system implementation

ejgallego:
> > * there is no possibility to systematically explore the list of objects currently in scope (other than hoping the ad-hoc internal   structures are somehow exposed)

For this see for example https://github.com/coq/coq/pull/16261 , and how the API with the reverse map introduced here is used in Search and Locate. If we move `Category.t` in that PR to the data, one could have a reverse map on categories too. What metadata to add is a bit open, for now I was thinking of documentation as sketched in the proposal for liberabaci.

> * in some cases (such as printing), we need to perform a costly (and messy) global imperative state update as to simulate that a module has been opened, due to the imperative nature of the module system implementation

Hopefully the Named trait could be used to avoid the global state update for the Nametab in `printmod.ml`

text/libobject-refactor.md
```diff
@@ -0,0 +1,228 @@
...
+
+Another interesting topic is versioning of objects. While preliminary,
+we could include versioning information in the objects as for example
+to allow to have hybrid Coq 8/9 objects, or several versions of the
+same `Definition` for example.
+
+# Part II: behavior specification
+
+The second part of this CEP, not yet implemented, is to phase out the
+`declare_object` API in favor a set of `Behavior.t` traits that an
```
maximedenes, line 167: Is there a description of the `Behavior` API?

ejgallego: I don't have a too precise view on how to implement it, there are several options so here we can use some feedback.

I guess for the first prototype I'd try to have a list of traits, indeed due to the existential nature of traits that would mean turning the declaration of objects into a module, but I think I'm fine with that (as in the case for errors)

ejgallego:
Ran out of time to work on this.

PR #67, opened 2023-01-04
[CEP] Sort variables in unification.

ppedrot:
Rendered: https://github.com/ppedrot/ceps/blob/sort-variable-unification/text/000-sort-variable-unification.md

Partially implemented as https://github.com/coq/coq/pull/16903.

### DIFF
--- a/text/000-sort-variable-unification.md
+++ b/text/000-sort-variable-unification.md
@@ -1,0 +1,77 @@
+- Title: Introducing quality variables in sort unification
+
+- Drivers: Pierre-Marie Pédrot
+
+----
+
+# Summary
+
+In addition to variables for universe levels, we generalize the type of sorts to
+also contain quality variables, allowing to delay the choice between Prop, Type
+and SProp made by the unification. This CEP is only concerned about unification,
+although the kernel structures need to be modified accordingly, the kernel
+treats undefined quality variables as an error, similarly to evars.
+
+The first part of this CEP is implemented in [Coq PR 16903](https://github.com/coq/coq/pull/16903).
+
+# Motivation
+
+There are several underlying motivations for this CEP.
+
+- Immediate fix: alleviate the longstanding issue of having to explicitly
+annotate variables with `Prop` to make the unification go through.
+- Short term: fix the `SProp` mess in the implementation of kernel and unification.
+- Long term: enhance the modularity of Coq by allowing terms to be polymorphic
+over sorts like `Prop` / `Type` / `SProp`, a.k.a. stop triplicating everything.
+
+The first point is adressed directly in this CEP, the two other ones are made
+possible after the CEP is implemented. We detail the issues below.
+
+As of Coq 8.17, every time one creates a fresh sort, it is implicitly living in the `Type` hierarchy. This has unfortunate consequences, like the fact that `fun A => A /\ A -> A` typechecks but `fun A => A -> A /\ A` does not. Since the arrow is typechecked left-to-right, the first one forces `A : Prop` but the second one forces `A` to be a fresh sort, hence a `Type`. In that latter case you have to explicitly write the Prop annotation for the pretyper to be happy. Similar issues arise from other asymmetrical constructs like equality.
+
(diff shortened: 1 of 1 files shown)


text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
+- Title: Introducing quality variables in sort unification
+
+- Drivers: Pierre-Marie Pédrot
+
+----
+
+# Summary
+
+In addition to variables for universe levels, we generalize the type of sorts to
+also contain quality variables, allowing to delay the choice between Prop, Type
```
JasonGross, line 10: Minor English grammar nit:
```suggestion
also contain quality variables, which allows delaying the choice between Prop, Type
```
or
```suggestion
also contain quality variables, allowing delaying the choice between Prop, Type
```
which reads a bit more awkwardly to me, but is more correct than "allowing to delay"

text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
...
+and SProp made by the unification. This CEP is only concerned about unification,
```
JasonGross, line 11: Minor nit:
```suggestion
and SProp made by unification. This CEP is only concerned about unification,
```
?

text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Motivation
+
+There are several underlying motivations for this CEP.
+
+- Immediate fix: alleviate the longstanding issue of having to explicitly
+annotate variables with `Prop` to make the unification go through.
+- Short term: fix the `SProp` mess in the implementation of kernel and unification.
+- Long term: enhance the modularity of Coq by allowing terms to be polymorphic
+over sorts like `Prop` / `Type` / `SProp`, a.k.a. stop triplicating everything.
```
JasonGross, line 25: Will this also finally replace template polymorphism, when combined with universe polymorphism?

JasonGross (commented):
Some minor suggestions

text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
...
+treats undefined quality variables as an error, similarly to evars.
+
+The first part of this CEP is implemented in [Coq PR 16903](https://github.com/coq/coq/pull/16903).
+
+# Motivation
+
+There are several underlying motivations for this CEP.
+
+- Immediate fix: alleviate the longstanding issue of having to explicitly
+annotate variables with `Prop` to make the unification go through.
```
JasonGross, line 22: Is this going to force `Prop` anytime you accidentally or intentionally use impredicativity?  e.g., is `Check let S := _ in let T : S := _ in (forall A : S, T) : S.` expected to succeed, instantiating `?S` to `Prop`?

text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Detailed design
+
+This CEP introduces a trivial idea to sort out the mess. Namely, in addition to the well-known universe *levels* one now also has access to sorts *qualities*. While levels are basically integers, qualities are the nature of the sort itself, i.e. `Prop`, `Type` or `SProp`. Since we want to have unification on this first-order signature, we also freely add variables. They are just tracked in the evarmap in a way similar to levels. The most generic type is thus `QSort (α, u)` where `α` is a sort quality and `u` an algebraic universe.
+
+Since this affects pervasive data structures, we need to change the kernel definitions. But similarly to evars, the kernel currently does not know anything about undefined sort variables, apart for their relevance which is needed for conversion.
```
JasonGross, line 38: Nit:
```suggestion
Since this affects pervasive data structures, we need to change the kernel definitions. But similarly to evars, the kernel currently does not know anything about undefined sort variables, apart from their relevance which is needed for conversion.
```

text/000-sort-variable-unification.md
```diff
@@ -0,0 +1,77 @@
...
+
+In order for the system to work, we have constraints and quotients going on.
+- The type of a sort `QSort (α, u)` is always `Type(u+1)`, i.e. the sort of a sort must have a `Type` quality. This is true in current Coq since `Prop : Type`, `SProp : Type` and `Type : Type`.
+- The type of a product `forall x : A, B : QSort(α, u)` is such that `α` is the quality of the sort of `B`, and u the max of levels (as usual). This is also true in Coq, and is designed to make impredicativity work.
+- Impredicative sorts have a degenerate level structure, i.e. `QSort(Prop, u) ≡ Prop` and similarly for `SProp`. This ensures that we can make unification of sort qualities commute with impredicative typing.
```
JasonGross, line 43: Does this play poorly with `QSort (α, u) : Type(u+1)`?  i.e., we can get `Prop : Type(Set+1)`, so does the unification engine know that `QSort (α, u) : Type(Set+1)` imposes a constraint on either `α` or `u`, but not necessarily both?

text/000-sort-variable-unification.md
ppedrot: No, both for usability and completeness. There are several impredicative universes so there is no reason to pick one randomly.

text/000-sort-variable-unification.md
ppedrot: Yes.

text/000-sort-variable-unification.md
ppedrot: This is indeed incomplete, but sound. Universe minimization should be able to erase those spurious constraints if they don't refer to universes that survive it.

ppedrot:
I think that the CEP is now implemented in master, so it should be merged.

Zimmi48:
Seems logical. Feel free to do so yourself. (I haven't followed the CEP or the PR, so it would seem strange for me to do it.)

PR #70, opened 2023-07-18
Draft rewrite rules for 069-coq-roadmap.md

TheoWinterhalter:
I'm not entirely sure about style so I did something. Please tell me whether this is what you expected or not.

### DIFF
--- a/text/069-coq-roadmap.md
+++ b/text/069-coq-roadmap.md
@@ -89,6 +89,10 @@ to add and remove items, to reflect the evolution of the project.
 
 #### Rewrite rules
 
+The goal is to add (unsafe) user-defined rewrite rules. It would be deactivated by default and be optionally on by *eg* setting `Rewrite Rules On`. They should supported by unification and the main reduction machines (not `vm_compute` and `native_compute` for now, Coq should give a warning when those are used with rewrite rules on). Rewrite rules would propagate to any module that requires the module that defines them.
+
+- Working Coq fork: https://github.com/Yann-Leray/coq (examples in https://github.com/Yann-Leray/coq/blob/rewrite-rules/test-suite/success/rewrule.v).
+- CEP: https://github.com/coq/ceps/pull/50
 - Yann Leray, Théo Winterhalter
 - 3 to 6 month
 
--- a/text/069-coq-roadmap.md
+++ b/text/069-coq-roadmap.md
@@ -89,7 +89,9 @@ to add and remove items, to reflect the evolution of the project.
 
 #### Rewrite rules
 
-The goal is to add (unsafe) user-defined rewrite rules. It would be deactivated by default and be optionally on by *eg* setting `Rewrite Rules On`. They should supported by unification and the main reduction machines (not `vm_compute` and `native_compute` for now, Coq should give a warning when those are used with rewrite rules on). Rewrite rules would propagate to any module that requires the module that defines them.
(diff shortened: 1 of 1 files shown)


TheoWinterhalter:
I tried to add a bit of that. :)

text/069-coq-roadmap.md
```diff
@@ -89,6 +89,12 @@ to add and remove items, to reflect the evolution of the project.
 
 #### Rewrite rules
 
+The goal is to add (unsafe) user-defined rewrite rules. This features allows users to add computation rules to axioms which can be useful for prototyping. It also allows for different kinds of computation rules with respect to what Coq currently permits: non-linearity, overlapping left-hand sides (*eg* one can write an addition on natural numbers that reduces on both sides: `0 + n` and `n + 0` both reducing to `n`).
+
+It would be deactivated by default and be optionally on by *eg* setting `Rewrite Rules On`. They should supported by unification and the main reduction machines (not `vm_compute` and `native_compute` for now, Coq should give a warning when those are used with rewrite rules on). Rewrite rules would propagate to any module that requires the module that defines them.
```
Zimmi48, line 94:
```suggestion
It would be deactivated by default and be optionally on by *eg* setting `Rewrite Rules On`. They should be supported by unification and the main reduction machines (not `vm_compute` and `native_compute` for now, Coq should give a warning when those are used with rewrite rules on). Rewrite rules would propagate to any module that requires the module that defines them.
```

Zimmi48:
Thanks!

text/069-coq-roadmap.md
```diff
@@ -89,6 +89,12 @@ to add and remove items, to reflect the evolution of the project.
 
 #### Rewrite rules
 
+The goal is to add (unsafe) user-defined rewrite rules. This features allows users to add computation rules to axioms which can be useful for prototyping. It also allows for different kinds of computation rules with respect to what Coq currently permits: non-linearity, overlapping left-hand sides (*eg* one can write an addition on natural numbers that reduces on both sides: `0 + n` and `n + 0` both reducing to `n`).
```
jfehrle, line 92: "prototyping" isn't mentioned in the CEP (#50).  It suggests that there are later steps to make something "final" or "production".  Perhaps you could explain what you have in mind in the CEP with a brief explanation of those later steps?

What fraction of our users will be willing/able to use this feature for important projects if it's unsafe?

I assume the prototyping is meant for users and not just Coq developers.

jfehrle:
@Zimmi48 FWIW, the rendering where you quote me above omitted "make it possible/much easier to do **\<some technical thing>**".  This '<' needed to have a backslash escape :-(

text/069-coq-roadmap.md
TheoWinterhalter: I don't understand why it should be mentioned in the CEP. These are just examples of what people want to do with rewrite rules: in particular we already have people using the fork of @yannl35133 to try out various type theories that are not otherwise encodable in Coq.

This use case has nothing to do with the status of rewrite rules.

It is true however, that the first instance we propose is something that is unsafe—in the sense that Coq may not terminate, or be incomplete for conversion if the user did not choose the right rewrite rules—and that the user must know that. To me this is the same as unsetting guard checking or using `Type : Type`, these are features used by actual people in actual projects even though they are not *safe*.

jfehrle: It's that "prototyping" is unclear.  As I said:

> It suggests that there are later steps to make something "final" or "production".

What are those steps?  Or is it a misleading choice of word?  Can you clarify?

TheoWinterhalter: I think there is a misunderstanding. A user may want to use rewrite rules for prototyping before they actually implement something, but it's up to the user so *I* don't have to explain the steps for this hypothetical user work.

PR #71, opened 2023-07-22
Towards a more flexible and accurate model for grammar levels and associativity

herbelin:
Rendered: https://github.com/coq/ceps/blob/camlp5-associativity/text/camlp5-associativity.md

### DIFF
--- a/text/camlp5-associativity.md
+++ b/text/camlp5-associativity.md
@@ -1,0 +1,146 @@
+- Title: Towards a more flexible and accurate model for grammar levels and associativity
+
+- Drivers: Hugo Herbelin
+
+----
+
+# Summary
+
+The Camlp5/gramlib model for associativity of levels in extensible grammar has a couple of issues. We reflect how to solve them.
+
+# Detailed description of the current problems
+
+## Non associativity
+
+Camlp5 let us believe that non associativity is possible and we let Coq users believe in turn that it was the case too. However, non associativity is *not* a feature of Camlp5. As said in the section [Associativity](https://camlp5.readthedocs.io/en/latest/grammars.html?highlight=associativity#extensible-grammars) of the Camlp5 manual, the keyword `NONA` behaves as `LEFTA`. And indeed, there is no mechanism to support nor emulate non associativity of infix or postfix operators in Camlp5.
+
+Also, non associativity (equivalently right associativity) is inactive on postfix operators (e.g., a rule such as `[ RIGHTA [ t = SELF; "!" -> ... ] ]` accepts parsing `x!!`. This is for the same reason as non-associativity is not supported (one would need to call continuations only once and not arbitrarily recursively many often, see remark at end of camlp5/camlp5#100).
+
+For another reason, non associativity (equivalently left associativity) is not emulable on prefix operators (e.g., a rule such as `[ LEFTA [ "?"; t = SELF -> ... ] ]`, or `[ LEFTA [ "?"; t = NEXT -> ... ] ]`, accepts parsing `??x`); this times it is because of a fallback which replace `SELF` by `SELF TOP` if `SELF` has failed.
+
+Note that the non support for non associativity in the parser is currently a cause of asymmetry with the printer, since the printer respects non associativity.
+
+## Each Camlp5 level is in practice two levels
+
+Indeed, a Camlp5 level is made of two blocks:
+- a block of rules of the form `SELF; suffix`
+- the block of rules not of this form
+
+In the first block, `SELF` is implemented by a call to the 2nd block of rules. As a consequence, if an infix notation is put at Coq's lower level 0, it still succeeds to parse its argument by finding closed subterms in the second block of level 0, such as an ident, or `( term )`.
+
+This results in bug reports such as coq/coq#15336 where the parser parses the notation but the printer does not print it.
+
+## Mixing associativity in the same level is technically possible and would solve the issue of combining developments that set different associativities to the same level
+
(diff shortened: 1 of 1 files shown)


text/camlp5-associativity.md
```diff
@@ -0,0 +1,146 @@
...
+## Removal of the recovery mechanism
+
+Removing the recovery mechanism breaks quite common situations such as `True \/ forall x, x = 0`, or `2^-n`, or `x~0%Z`.
+
+One could try to automatically adapt the scripts but I believe that the simplest is to move `forall` (and all `binder_constr` rules spanning until the end of the sentence) to a lower level, such as 10, just above the level of application.
+
+The drawback is that we need to modify the printer so that it knows how to print such cycle made of a rule at some level (here 10) which is lower than the level on the right-hand side of the rule (here 200). This is however already made possible by PR #17875.
+
+For `-`, there are several solutions, such as:
+- to set it at level 10 with arguments at level 40 (the level of `*`) so that `2 ^ - x * y` is `2 ^ - (x * y)` but `2 ^ - x + y` is `(2 ^ - x) + y` (??)
```
silene, line 76: Parsing `2 ^ - x * y` as `2 ^ (- (x * y))` seems to be quite surprising. I would expect the consensus among mathematicians to be `(2 ^ (- x)) * y`.

In other words, there is no reason for the argument of prefix `"-"` to be at level 40. It should be at the current level (i.e., 35 if the argument of infix `"^"`. In fact, the notion of levels for a notation that starts with a terminal and ends with a non-terminal does not make sense in the first place. It should not have an intrinsic level, it should always use the ambient level.

herbelin: I got confused by the idea that `- x * y` was `- (x * y)` by default  in mathematics, but that's the same as `(- x) * y`, so it does not have to be the former. And actually, most programming languages seem to make prefix `-` binding tightly (so morally at 10 for Coq).

For the example I mentioned, the level of `x` in `- x` do not matter. The whole notation could be at 10 with `x` at 10 too.

What do you mean by "It should not have an intrinsic level, it should always use the ambient level.". That is, in the following examples, how would you associate (or fail):
- `x * - y `
- `x ^ - y`
- `x + - y`
- `f x - y`
- `- f * x`
- `- f  ^ x`
- `- f + x`
- `- f x`

silene: > And actually, most programming languages seem to make prefix `-` binding tightly (so morally at 10 for Coq).

That does not mean that it is a good idea. In fact, the very first sentence of the Wikipedia page you cite says that `- 3 ^ 2` is interpreted as `-9` in mathematics. And the only "languages" it mentions as binding tightly are Excel, PlanMaker, and bc.

So, having unary minus at level 10 would be awful. And having it at level 35 is already painful. I hate having to write `- (x * y)`. It should have been at level 45.

> * `x * - y`
> * `x ^ - y`
> * `x + - y`

All three of them associate as `x ! (- y)`.

> * `f x - y`

Do you have an infix minus in the context? If not, then parsing can be recovered as `f x (- y)`.

But the interesting case is actually `f x - y z`, which would be parsed as `f x (- y) z` (assuming again that there is no infix minus, otherwise the point is moot).

> * `- f * x`
> * `- f  ^ x`
> * `- f + x`
> * `- f x`

You did not say what the ambient level is, so it depends. Let us assume that those expressions all look like `y + - f ! x` (so, ambient level is 50 left associative). Then they would respectively parse as

* `y + (- (f * x))`
* `y + (- (f ^ x))`
* `(y + (- f)) + x`
* `y + (- (f x))`

herbelin: (I'm tired, the first 3 questions were pointless. I meant the minus sign in the front)

I don't see what conclusion to draw otherwise. And we are somehow constrained by what is currently done, which is:
```coq
Require Import ZArith. Context (x y :Z) (f : Z -> Z) (g : Z). Set Printing Parentheses. Open Scope Z_scope.
Check  x * - f y. (* x * (- (f y)) *) (* due to recovery *)
Check  x ^ - f y. (* x ^ (- (f y)) *) (* due to recovery *)
Check  x + - f y. (* x + (- (f  y) *) (* due to recovery *)
Check  f x - y.   (* (f  x) - y *)
Check  - g * x.   (* (- g) * x *)
Check  - g ^ x.   (* - (g ^  x) *)
Check  - g + x.   (* (- g) ^  x *)
Check  - f x.     (* - (f x) *)
Check  - x * y.   (* (- x) * y *)
Check  - x ^ y.   (* - (x ^ y) *)
Check  - x + y.   (* (- x) + y *)
```

herbelin: For the record, experimenting changing associativity of:
- `- x * y` from `(- x) * y` to `- (x * y)` and
- `- x / y` from `(- x) / y` to `- (x / y)`

requires several changes in stdlib, starting with:
- `Ring_theory.Ropp_mul_l : -(x * y) == -x * y` (also in `Ncring.v`) and `Ring_theory.ARopp_mul_l : -(x * y) == (-x) * y`
- `BinInt.Zopp_mult_distr_l : - (n * m) = - n * m`
- `Zdivfloor.div_opp_opp : b~=0 -> -a/-b == a/b`
- `Field_theory.rdiv5 : - (a / b) == - a / b`
- `Uint63.opp_spec : φ (- x) = - φ x mod wB`
- `-a÷b` in `Zquote.v`
- `- Qnum x * Zpos (Qden y)` in `QArith_base.v`
- non exhaustive list...

PR #72, opened 2023-07-31
Discharge on the fly

herbelin:
Rendered [here](https://github.com/herbelin/ceps/blob/discharge-on-the-fly/text/discharge-on-the-fly.md).

### DIFF
--- a/text/discharge-on-the-fly.md
+++ b/text/discharge-on-the-fly.md
@@ -1,0 +1,109 @@
+- Title: Discharge on the fly
+
+- Drivers: Hugo Herbelin
+
+----
+
+# Summary
+
+After the refactoring of discharge made in coq/coq#14727, it becomes possible to discharge declarations at the time of their declaration. We discuss a proposal in this direction.
+
+# Motivation
+
+It is common in a section to want to temporarily generalize the notions defined in the section but to continue to work in the context of the section for further developments, as in:
+
+```coq
+Section S.
+Context {A B:Type}.
+Inductive option := None | Some (a:A).
+Definition map (f : A -> B) (o : option) :=
+  match o with
+  | None => Top.None
+  | Some a => Top.Some (f a)
+  end.
+(* ... *)
+End S.
+```
+
+This is what the proposal tends to achieve.
+
+# Details about the design
+
+## Basic idea
+
+The basic proposal is to emulate a section
+```coq
+Section S.
+Variable a : A.
+Definition f := ...
+```
+as
+```coq
+Definition f a := ...
+Module S.
+Parameter a : A.
+Definition f := ...
+```
+
+that is, that each persistent declaration of the section is declared both in the section and, discharged, at toplevel (as well as appropriately discharged in all intermediary section levels).
+
+The resulting full names are typically `Top.f` for the toplevel copy, and `Top.S.f` for the copy in the section (as well as `Top.S.T.f`, etc., if there are nested sections).
+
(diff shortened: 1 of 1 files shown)


jfehrle:
This PR includes a second CEP that you proposed a week or two ago.  Is that what you intended?

herbelin:
Not intended :(.Sorry, removed.

text/discharge-on-the-fly.md
```diff
@@ -0,0 +1,109 @@
...
+The resulting full names are typically `Top.f` for the toplevel copy, and `Top.S.f` for the copy in the section (as well as `Top.S.T.f`, etc., if there are nested sections).
+
+At the end of the section, the module being built is discarded.
+
+The distributivity of declarations over section levels is obtained by maintaining one environment/state for each section level:
+- opening a section forks a new copy of the current environment/state for the section; this copy becomes the active one
+- extending a section content enriches simultaneously and hereditarily all section levels (i.e. `Definition f := t` adds `Top.f` to all section levels, `Top.S.f` to all section levels from level `Top.S`, `Top.S.T.f` to all section levels from level `Top.S.T`, etc.); there are however three kinds of declarations (see below)
+- closing a section drops the active copy of the environment/state and makes active the environment/state of the outer section (which itself growed since the fork)
+
+## Definitionally connecting the different copies
```
CohenCyril, line 59: Definitional equality is not enough in practice, the various copies must also be identified in the following contexts:
- in the syntactic phase of rewrite pattern matching
- in patterns of `Hint`
- as canonical keys
- in `Search` results

(and I'm not sure how exhaustive is this list)

There are two ways to address this problem:
1. Use a notation instead of a definition for aliases in each sections
2. Call [`Declare Equivalent Keys`](https://coq.inria.fr/distrib/current/refman/proofs/writing-proofs/equality.html?highlight=equivalent%20keys#coq:cmd.Declare-Equivalent-Keys) all the time and make sure it is taken into account by every relevant part of the system (and document it properly cf https://github.com/coq/coq/issues/4551).

Ideally, intermediate definitions should be discarded when exiting the section.

(Optionally, also, in the future, it would be nice if re-entering a named section could provide the same context and the same aliases....)

herbelin: > as canonical keys

In my experiences to provide simultaneously the non-discharged and discharged versions of a canonical key, I had several incompatibilities. Studying the reason of these incompatibilities might help to understand if there is a way to make it working, but that does not seem to be straightforward to support (see for instance the [failure](https://github.com/coq/coq/pull/17888#issuecomment-1656821741) on mathcomp I got when the discharged versions of a canonical projections are also allowed to be used).

> the various copies must also be identified

By identified, I guess you mean `eq_constr`-equal (alpha-convertible) vs `conv`-equal. Is it that you think that it would help compatibility? In any case, I struggled in not breaking too much compatibility in coq/coq#17888. I'm a bit pessimistic about the room for manœuver that we have.

> (Optionally, also, in the future, it would be nice if re-entering a named section could provide the same context and the same aliases....)

I agree. And it is not impossible to do if we accept to keep the inside of sections in vo files. And, at worst, even if keeping the inner of sections in vo files happens to be too costly in size, we may also recompute dynamically the instances.

CohenCyril:
Regarding the drop of `parameter` in Lean4 here is a story:
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/parameter.20vs.20variable/near/391936460

PR #78, opened 2023-10-05
Add Ltac* debugger and GUI support items to roadmap

jfehrle:


text/069-coq-roadmap.md
```diff
@@ -171,6 +171,41 @@ Would require work on:
...
+
+A number of GUI enhancements require support from Coq.  Given that
+vsCoq2, Proof General and coqtail are written in different languages,
+these should be made available through new calls in the GUI protocol
+to maximize the potential for code reuse.
+
+These include:
+- Data on where symbols are defined and used so users can jump to a symbol definition, etc.
+- Incrementally rebuilding the project (identifying what relevant files have changed)
+- Parsing the _CoqProject file should be done in Coq, then the result should be returned
```
ejgallego, line 205: This is actually not accurate, quite a few tools use the OCaml parsing library we have.

text/069-coq-roadmap.md
```diff
@@ -171,6 +171,41 @@ Would require work on:
...
+
+### Enabling GUI enhancements for vsCoq2, Proof General, etc.
+
+- developers TBD
+
+A number of GUI enhancements require support from Coq.  Given that
+vsCoq2, Proof General and coqtail are written in different languages,
+these should be made available through new calls in the GUI protocol
+to maximize the potential for code reuse.
+
```
ejgallego, line 201: I think no new calls should be added to the GUI protocols for the below items, in some cases because the functionality is already there, in some others because they are not needed.

text/069-coq-roadmap.md
```diff
@@ -171,6 +171,41 @@ Would require work on:
 - Module dependency analysis
 - Hugo Herbelin + Yannick Forster
 
+### Ltac* debugger
+
+#### Ltac2 support and enhancements
```
ejgallego, line 176: It is good seeing further debugger progress, but I think the workplan here should be very different from the way things were done in Ltac1:
- core debugging support / API should be added to Ltac2, this API is IDE / protocol-independent and should be discussed with Ltac2 developers; there are some API proposals but this is in general a challenguing area
- communication of the debugger with IDEs should be first studied as being implemented using the DAP protocol; if DAP doesn't work we should understand why and what the changes are needed.

jfehrle: > It is good seeing further debugger progress

I don't think you're sincere about this.  Originally I planned to make incremental improvements, but you took forever to review the first improvement, which I submitted 20 months ago.  You don't seem to be focused on delivering useful functionality in a reasonable timeframe.

I would say the train left the station on your suggestions.

text/069-coq-roadmap.md
jfehrle: You think there shouldn't be a category for Coq improvements to facilitate GUI functionality?
These are examples to illustrate the concept.  I could add a disclaimer saying that.  These are not even things I plan to work on personally.

ejgallego: I think there should be a category for Coq UI of course. What I mean is that the current active protocols we have already support such functionality, so update to the protocol is needed.

text/069-coq-roadmap.md
ejgallego: I am sincere about wanting to see further debugging support, and in fact supporting the Ltac debugger is on coq-lsp roadmap.

I am also sincere about my concerns about the technical and methodological problems I have found in the original proposal, which led to merging buggy code to Coq that took quite a while for me (and others to repair back), I am also concerned about protocol design methodology, and that's not an arbitrary concern as maintenance of interfaces is usually the number 1 problem interface code face.

Note that at least an additional CoqIDE maintainer does agree with my technical and social analysis, if my understanding is correct.

We are very happy to get contributions, however it is also important that our feedback about Coq and OCaml best practices is taken into account by the contributors, in particular when in this case it is the first time they write OCaml code.

I don't have an exact recollection of the circumtansces of the review of that first update, but indeed I can recall having not a good reviewer/coder communication, and indeed, as my suggestions were not taken into account or only did so [pretty reluctantly](https://github.com/coq/coq/pull/13783#discussion_r605190547) it is likely my motivation may not have been the best.

I am sorry the review experience wasn't a positive one for neither of us, and I'll be happy to discuss how to improve the contributing experience for all.

I would start by asking you to change the tone of your messages as in particular I'd like to remind that we have a Code of Conduct which I'm unsure your message above fully adheres to.

I would also like to request you to stop stating misleading or wrong information about the IDE work. I'm starting to feel this has become a pattern which is not good for anyone in the project.

> You don't seem to be focused on delivering useful functionality in a reasonable timeframe.

That's what you think, but my track record speaks otherwise, having started, maintained, and delivered multiple working prototypes over the years in different domains, including large Coq projects.

I would rather say I am focused on delivering **bug free* functionality that can be properly maintained.

> I would say the train left the station on your suggestions.

I am not sure I understand what that means, can you clarify?

jfehrle: > I am also sincere about my concerns about the technical and methodological problems I have found in the original proposal, which led to merging buggy code to Coq that took quite a while for me (and others) to repair back.

Neither the bugs (some oversights) nor relevant "technical and methodological" were identified as part of the review.  It took you and others time to fix them only because you and others chose to ignore several PRs I submitted with fixes.  I'm accustomed to fixing bugs that I caused.  And IIRC you submitted several bad fixes for one of the bugs--which I pointed out.

> it is also important that our feedback about Coq and OCaml best practices is taken into account by the contributors, **in particular when in this case it is the first time they write OCaml code**.

Perhaps for the sake of emphasis, this departs from reality.  My first OCaml project was proof diffs, about 5 years ago, which you reviewed.  Remember?

> I am sorry the review experience wasn't a positive one for neither of us, and I'll be happy to discuss how to improve the contributing experience for all, as indeed I am convinced we can do better in many fronts.

Agreed, that would be a good discussion for the group.

> I would also like to request you to stop stating misleading or wrong information about the IDE work.

Perhaps you would educate me on what's misleading or wrong.  What I suggested here is a strawman.  Where does one find the correct information, preferably in the form of a document?

>> You don't seem to be focused on delivering useful functionality in a reasonable timeframe.

> That's what you think, but my track record speaks otherwise, having started, maintained, and delivered multiple working prototypes over the years in different domains, including large Coq projects.

FTR, I was commenting on your ability to work well with others, to be a good team player, not on your track record.

>> I would say the train left the station on your suggestions.

I have almost completed my prototype, so suggestions about how to plan the work are not timely.  Of course there will be discussion and review of the prototype when it is ready.  There's no point in adding DAP support to the CoqIDE interface, but it is an open and timely question for the "Debugger support in vsCoq2" item.

PR #79, opened 2023-10-10
Update 069-coq-roadmap.md

yforster:


Zimmi48:
Someone said in the Coq Call yesterday that this should be clarified that certified extraction will be an alternative proposed to the user rather than a drop-in replacement.

@mattam82 also seemed to think that some text written there was lost. Is that the case?

In any case, please let me know when this is ready to merge, or @mattam82 can merge directly.

PR #90, opened 2024-05-21
Template poly redesign using sort poly

SkySkimmer:
rendered https://github.com/coq/ceps/blob/template-poly/text/XXX-template-poly.md

### DIFF
--- a/text/XXX-template-poly.md
+++ b/text/XXX-template-poly.md
@@ -63,6 +63,8 @@ We then check any constraints on the variables and do regular typechecking.
 
 ## Subject reduction
 
+### Inductive
+
 Consider `(fun X:Type@{u} => I X) (P:Type@{v}) (Q:Type@{w})`
 with `I@{i j | csts(i,j)} (p:Type@{i}) (q:Type@{j}) : Type@{f(i,j)}`
 and default instance `{i0 j0}` (which must verify `|= csts(i0, j0)`).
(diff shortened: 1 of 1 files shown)


text/XXX-template-poly.md
```diff
@@ -0,0 +1,154 @@
...
+## Note on nested templates and above Prop
+
+Consider
+~~~
+Inductive double@{q|u|} (A:Type@{q|u}) : Type@{q|u}
+  := Double : prod A A -> double A.
+~~~
+When checking `prod A A` we must ensure that `q` is above Prop,
+and if we had `prod A B` with A and B at different quality variables
+we would have to unify them (in elaboration) or error (in kernel).
```
SkySkimmer, line 102: I guess `prod A B` with A and B at different qualities would land in Type thanks to above Prop

SkySkimmer: but elaboration still need to have some unification otherwise `Inductive foo A B := wrap (_ : prod A B)` will produce `foo : _ -> _ -> Type@{Type|_}` I think

text/XXX-template-poly.md
```diff
@@ -0,0 +1,154 @@
...
+The quality (if bound) must be "above Prop".
+
+If the quality is bound (ie non constant), it is assigned to the max of the inferred qualities
+(this is possible because of the "above Prop" assumption).
+For each bound univ level, it is assigned to the max of the corresponding obtained levels.
+
+This gives the implicit instance.
+We then check any constraints on the variables and do regular typechecking.
+
+## Subject reduction
```
SkySkimmer, line 64: TODO check that no problem appears with constructors

text/XXX-template-poly.md
```diff
@@ -0,0 +1,154 @@
...
+  If it has one, it is the quality of the output sort.
+- each parameter type either does not mention the bound univs (CLOSED),
+  or is of the form `forall args, Type@{q|u}` (BINDING)
+  where the `args` types do not mention the bound univs and qvar,
+  `q` is either Type or the unique bound qvar,
+  and `u` is either constant or one of the bound univ levels.
+- the indices types and constructor types do not mention the bound univs and qvar.
+
+NB: the qvar (if there is one) must be "above prop" so should not appear in relevance marks
+(it's always `Relevant`) so eg
```
SkySkimmer, line 36: not fully clear if we want to keep the above_prop restriction
it makes this relevance issue easier but at the cost of needing the kernel to understand above_prop

text/XXX-template-poly.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Detailed design
+
+## Declaration
+
+An inductive may be declared to use implicit univ instances when:
+
+- it is non mutual (nested is ok) (this rule is in master but maybe we will remove it)
+- its univ declaration has only univ variables unbounded from below
+- its univ declaration has at most 1 qvar variable.
```
SkySkimmer, line 26: perhaps it would make sense to relax this condition, the extra qvars would appear in the parameters but not in the return sort. if we drop the above_prop condition it may be mildly useful, eg could define a template poly
~~~coq
Class foo A B := {}.
~~~
which can accept both SProp and Type for A and B

text/XXX-template-poly.md
```diff
@@ -0,0 +1,154 @@
...
+  and `u` is either constant or one of the bound univ levels.
+- the indices types and constructor types do not mention the bound univs and qvar.
+
+NB: the qvar (if there is one) must be "above prop" so should not appear in relevance marks
+(it's always `Relevant`) so eg
+~~~
+Inductive prod@{q|u v|} (A:Type@{q|u}) (B:Type@{q|v}) : Type@{q|max(u,v)} :=
+  pair : A -> B -> prod A B.
+~~~
+is accepted and the arrow `A -> ...` has relevance mark `Relevant` not `RelevanceVar q`.
```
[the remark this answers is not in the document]
SkySkimmer: so that instantiation is trivial, but this is probably not actually important

text/XXX-template-poly.md
SkySkimmer: In fact we may not need to understand above_prop, we can allow template qualities to be instantiated by anything. 
However taking the max of the argument qualities to get the implicit instance may then fail.

This means for instance we can have `sigma@{q|a b} : forall (A:Type@{q|a}) (B:A -> Type@{q|b}), Type@{q|max(a,b)}` such that
`sigma True (fun _ => nat) : Set`
`sigma True (fun _ => True) : Prop`
`sigma sTrue (fun _ => sTrue) : SProp`
`sigma A B : Type@{q|...}` if A and B are in the same qvar `q`

and `sigma True (fun _ => sTrue)` and `sigma A B` with A and B at different qvars are errors.

text/XXX-template-poly.md
SkySkimmer: also matches

mattam82: A priori the same must be done for (partial applications of) constructors otherwise typechecking their parameter arguments them will naturaly add constraints on the default instance.

text/XXX-template-poly.md
```diff
@@ -0,0 +1,200 @@
...
+Consider `(fun X:Type@{u} => I X) (P:Type@{v})`
+with `I@{i | csts(i)} (p:Type@{i}) (q:Type@{i}) : Type@{f(i)}`
+and default instance `{i0}` (which must verify `|= csts(i0)`).
+
+It has type `forall Type@{max(u,i0)}, Type@{f(u,i0)}` assuming
+- `|= csts(u, j0)`
+- `|= v <= u`
+
+The reduced value `I P` has type `forall Type@{max(v,i0)}, Type@{f(v,i0)}`
+which is NOT comparable to `forall Type@{max(u,i0)}, Type@{f(u,i0)}`!!
```
SkySkimmer, line 106: This problem arises from non linearity of the template universes in the parameters.

Current plan:
- for qualities, we need nonlinearity as we can't put `max(q1,q2)` in the return type. So we fix the problem by saying that a partially applied template requires the quality to be Type.
- for levels we forbid nonlinearity (the alternative would be to say that we use the implicit instance for partially applied templates, and only get universes from arguments when fully applied; not fully clear which choice is better or if we should do both)

PR #93, opened 2024-10-04
Roadmap: attenuate the marketing tone slightly

gasche:
At the last [Coq call](https://github.com/coq/coq/wiki/Coq-Call-2024-10-01), @herbelin mentioned that the beginning of the Roadmap document feels a bit chatgpt-generated. The way I understand this comment is that the tone is trying to imitate marketing speech, which is uncommon in our community, and not necessarily something people are comfortable with -- clearly Hugo is not.

I think that it is good to be able to show, with a choice of tone: "we are serious about improving our communication, trying to become more popular". But hopefully it is possible to find a good middle ground that does not push people away by sounding like snake oil.

I'm sure that the present document is already a compromise reached by several people trying to find a good balance, but this discussion motivated me to try to tone the marketing down *just a little bit*.

This PR contains three independent commits, very small changes, only in the first few paragraphs:

- Rephrase the sentence that talks about "agile development" and "accelerating innovation", to say the same thing without those negatively-associated keywords.
- Rephrase the pillars to remove their fluffy adjectives.
- Rephrase the "Usable" description to not *start* with AI-powered tool. It's a bit unfair because I think that Coq is a place where applying AI makes sense, but personally I get bad vibes from projects that talk about AI first (or, worse, the blockchain), so I think that putting it as a second sentence could help.

PR #94, opened 2024-10-09
Propose an alternative for the second paragraph of the roadmap.

Zimmi48:
This is a different attempt at improving the paragraph that remains debated in #93 (the commits changing the rest of the introduction having been merged already). Based on discussions in this pull request, it tries to make it clearer what we meant by "leverages the rich ecosystem" and a "more agile development philosophy".

We keep the word "agile", but we remove the word "more" (as suggested by Hugo) and the word "philosophy" (as suggested by Erik). We remove the word "leverages" (as suggested by Jim) in favor of "builds upon". We insist on our ecosystem and our care for compatibility and ecosystem maintainability (which was only implicit so far). We add "research" (as suggested by Gabriel), but we keep "innovation" next to it to insist that we want to see impact both in research and in industry.

### DIFF
--- a/text/rocq-roadmap.md
+++ b/text/rocq-roadmap.md
@@ -5,7 +5,7 @@ Rocq Long Term Roadmap
 
 The Rocq Development Team, building on 40 years of experience with Coq, is pleased to announce the Rocq Prover, the next-generation proof assistant. The Rocq Prover is designed to empower a diverse range of users, from mathematicians seeking formal rigor to software engineers building high-assurance systems.
 
-While Rocq leverages the rich heritage of Coq, it embraces a more agile development philosophy. This allows us to deliver impactful changes more frequently, accelerating innovation and user experience improvements.
+Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a development methodology which achieves the delicate balance of being mindful of compatibility and ecosystem maintainability, while sufficiently agile to deliver impactful changes frequently, boosting research and innovation, and continuously improving user experience.
 
 This roadmap outlines our vision for Rocq's future, focusing on four key pillars:
 
--- a/text/rocq-roadmap.md
+++ b/text/rocq-roadmap.md
@@ -5,7 +5,7 @@ Rocq Long Term Roadmap
 
 The Rocq Development Team, building on 40 years of experience with Coq, is pleased to announce the Rocq Prover, the next-generation proof assistant. The Rocq Prover is designed to empower a diverse range of users, from mathematicians seeking formal rigor to software engineers building high-assurance systems.
 
-Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a development methodology which achieves the delicate balance of being mindful of compatibility and ecosystem maintainability, while sufficiently agile to deliver impactful changes frequently, boosting research and innovation, and continuously improving user experience.
+Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a fast-paced development methodology to deliver impactful changes frequently, supporting academic and industrial research, and continuously improving user experience. At the same time, we are mindful of compatibility and ecosystem maintainability.
 
 This roadmap outlines our vision for Rocq's future, focusing on four key pillars:
 


text/rocq-roadmap.md
```diff
@@ -5,7 +5,7 @@ Rocq Long Term Roadmap
 
 The Rocq Development Team, building on 40 years of experience with Coq, is pleased to announce the Rocq Prover, the next-generation proof assistant. The Rocq Prover is designed to empower a diverse range of users, from mathematicians seeking formal rigor to software engineers building high-assurance systems.
 
-While Rocq leverages the rich heritage of Coq, it embraces a more agile development philosophy. This allows us to deliver impactful changes more frequently, accelerating innovation and user experience improvements.
+Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a development methodology which achieves the delicate balance of being mindful of compatibility and ecosystem maintainability, while sufficiently agile to deliver impactful changes frequently, boosting research and innovation, and continuously improving user experience.
```
gasche, line 8: I would split this in shorter sentences.

> Rocq {builds upon, leverages} the rich heritage and diverse ecosystem of Coq; it is mindful of compatibility and ecosystem maintainability. It also keeps evolving; we deliver impactful changes frequently, practice active research, and continuously improve the user experience.

Zimmi48: So you still want to remove the word "agile" in this context?

gasche: "agile" and rings bullshit-bells to me; I think that you are less sensitive about it because you talk to software-engineering people, but 99% of the times I hear this word it is in the context of someone spewing nonsense. Same with "innovation". Personally I would avoid those words because of their largely negative connotations among researchers (which remain the main audience of Coq).

Zimmi48: Oh right, I had not noticed that you were also removing the word "innovation".

I think we are not going to converge on something that satisfies us both here, so maybe we should bring this topic back to the next Coq Call?

herbelin: Regarding the use of "agile" by @erikmd, it was to mention the advanced use of tooling and the regular developer meeting. This can probably be said explicitly some way or another without using the word "agile".

As for "innovation", isn't it already morally covered by the words "research", "impactful", "improve", "evolving"?

Of course, it depends also on the audiences we want to target.

jfehrle, line 8: For some reason, I can't use the suggestion widget on this text :-(.

> We promote a development methodology which achieves the delicate balance of being mindful of compatibility and ecosystem maintainability

A lot of vague words.  Maybe something like this, but later in the document:

"We aim to maintain backward compatibility for existing users to a large degree without unduly limiting our ability to deliver improvements.  You may need to make limited changes to migrate your project to future versions of Rocq."

> boosting research and innovation,

This is a top-level goal, it should be in its own sentence, perhaps in the first paragraph.  Maybe something like

"The Rocq Development Team works to support improved proof verification techniques, to promote greater use and understanding of proof verification and to support research in topics related to proof verification"

> while sufficiently agile to deliver impactful changes frequently,
> and continuously improving user experience.

These could apply to almost any software project and therefore they're not interesting.  They would be a lot more interesting if they provided new information tied to the specifics of Rocq.  But they probably shouldn't be in this position in the document.

text/rocq-roadmap.md
```diff
@@ -5,7 +5,7 @@ Rocq Long Term Roadmap
...
 
 This roadmap outlines our vision for Rocq's future, focusing on four key pillars:
```
jfehrle, line 10: At some point, would you split the long lines in this file?

jfehrle (commented):
My thoughts.

text/rocq-roadmap.md
```diff
@@ -5,7 +5,7 @@ Rocq Long Term Roadmap
 
 The Rocq Development Team, building on 40 years of experience with Coq, is pleased to announce the Rocq Prover, the next-generation proof assistant. The Rocq Prover is designed to empower a diverse range of users, from mathematicians seeking formal rigor to software engineers building high-assurance systems.
```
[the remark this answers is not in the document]
Zimmi48: I think we will keep the first paragraph as is, or very close to its current form, for the Rocq website launch. Then, we can consider updating it after a few months to be less out of date.

However, I'm strongly opposed to write "formerly known as Coq" anywhere in our communication material:

- if people were not aware of this, they will not care
- if people were already aware, they don't need it
- if people need to find what has become of Coq, they will quickly find it

Mentioning Coq is only worth when we want to highlight our experience with developing proof assistants. The way this is currently formulated does not suggest "old software". The fact that we work by incremental improvement is a development methodology, but it doesn't make Coq closer to what it was 40 years ago.

text/rocq-roadmap.md
Zimmi48: If you want this to happen, feel free to open a PR doing it. Personally, I don't quite see the benefit.

text/rocq-roadmap.md
Zimmi48:
```suggestion
Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a development methodology which achieves the delicate balance of being mindful of compatibility and ecosystem maintainability, while sufficiently fast-paced to deliver impactful changes frequently, boosting academic and industrial research, and continuously improving user experience.
```

Since so many people dislike the words "agile" and "innovation", this is my proposal to remove them without changing the meaning.

There was also an issue with a too-long sentence, so this could become:

```suggestion
Rocq builds upon the rich heritage and diverse ecosystem of Coq. We promote a fast-paced development methodology to deliver impactful changes frequently, supporting academic and industrial research, and continuously improving user experience. At the same time, we are mindful of compatibility and ecosystem maintainability.
```

I personally liked my first proposal better than the second one, and the second one better than the third one, but if there is a clear preference for any of these, I'm fine with them. If this is still not enough and people do not think this is a sufficient improvement on the current status, then I will not bother further and close this PR, to let others do their own attempts.

gasche: Yes, I do think there is a significant benefit in removing those loaded words to sound "less bullshit". Either proposals are fine, I think that the last one reads better due to having shorter sentences. Thanks!

text/rocq-roadmap.md
jfehrle: For one thing, it lets you comment on smaller parts of the text.  I think of it as standard procedure, just like for OCaml source code--really long lines are not as easy to read/work with.  But to each his own.

text/rocq-roadmap.md
jfehrle: > > These could apply to almost any software project and therefore they're not interesting.
> 
> That's not true, though. Lots of software aims for strong backward-compatibility at the detriment of fast-paced evolution.

My comment was about this text rather than backward compatibility:

> while sufficiently agile to deliver impactful changes frequently,
and continuously improving user experience.

"agile" is vague.  Maybe something about frequent releases would be better
"frequently" is in part (maybe in large part) a question of resources
"continuously improving user experience" - this seems like an aspirational goal.  At the least, I'm not sure what recent work I'd put under "improved user experience"

text/rocq-roadmap.md
Zimmi48: OK, then I will do it in a commit on master after this PR is merged (assuming there are no other PR open on this text at the same time, to avoid merge conflicts).

gasche: If I can hazard a comment: I think that it's mostly a tooling question (do your tools nudge you towards hard-wrapping or do they nudge you towards long line?), where older tools typically favored hard-wrapping to make the work of some line-oriented tools easier, while newer tools typically favor long lines for text documents (unstructured, unindented), for example Markdown.

In Github Markdown, and most Markdown processors these days, you are punished for hard-wrapping because it keeps the newlines in the rendered output (while the original markdown specification is design to consume hard-wrapped text nicely, by requesting two spaces at the end to force a newline). git-centric workflows are trying to work better with non-wrapped texts these days, for example some diff viewers will show you a word diff by default instead of a line diff, making any issue with long lines go away. On the other hand, hard-wrapped texts tend to be fairly unstable to reflows, and so they don't work so nice with line-based diff tools either.

Personally I keep hard-wrapping in Emacs out of habit, but I un-wrap my text before sending it to any web form. I don't have a strong preference for the wrapping strategy to use on versioned markdown documents.

text/rocq-roadmap.md
Zimmi48: > At the least, I'm not sure what recent work I'd put under "improved user experience"

Tons of work has recently been done for improved user experience. In fact, the part about usability in this document is becoming too long, precisely because there is so much going on there (most of the items listed are not just aspirational but correspond to real ongoing work).

text/rocq-roadmap.md
Zimmi48: > In Github Markdown, and most Markdown processors these days, you are punished for hard-wrapping because it keeps the newlines in the rendered output

I believe this is only the case for the markdown used in comments, but not for the markdown used in documents.

text/rocq-roadmap.md
jfehrle: "User experience" could be taken to encompass any user-visible change.  That definition makes the term useless.  I would limit the term to things that make it easier for users to do things that were already reasonably possible.  And often a fairly small, incremental changes and perhaps user interface-related.  Otherwise I think of them as new features.  It's a fuzzy line between them.  "Theory Transfer Support" sounds like a new feature, AFAIK there's nothing like that now.

Zimmi48: I agree, we should categorize these improvements better.

PR #105, opened 2025-02-13
Reduction

Janno:
Rendered: https://github.com/bluerock-io/ceps/blob/janno/reduction/text/XXX-reduction.md

Note that the document does not directly propose anything yet. It is intended to serve as a collection of existing challenges. The kind of examples we collect will hopefully guide us in figuring out what exactly we want to propose.

### DIFF
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -7,7 +7,7 @@
 | vm_compute    | ✅✅✅         | Full          | ❌                         | ❌                       | ❌                | ✅✅✅ Bespoke Cast honored by Qed                                                        |
 | lazy          | ✅✅           | Full / Head   | ❌                         | ✅✅                     | ❌                | ✅✅ Default cast does not store reduction flags                                          |
 | cbv,compute   | ✅             | Full / Head   | ❌                         | ❌                       | ❌                | ✅ Default cast coincides but will use kernel reduction/conversion instead of cbv         |
-| hnf           | ❔             | Head          |  ❔                        |  ❔                      |  ❔               |  ❔                                                                                       |
+| hnf           | ❔             | Head          |  ❔                        |  ❌                      | ✅                |  ❔                                                                                       |
 | simpl         | ⛔⛔           | Full / Head   | ✅/⛔                      | ⛔ (only head)           | ✅                | ⛔ Uses default cast                                                                      |
 | cbn           | ⛔⛔⛔         | Full / Head   | ✅✅                       | ✅✅                     | ✅✅              | ⛔ Uses default cast                                                                      |
 | **Proposals** |
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -4,15 +4,15 @@
 
 | Strategy      | Performance    | Normalization | Arguments                 | Reduction Flags          | Refolding        | Effect on Qed                                                                          |
 |---------------|----------------|---------------|---------------------------|--------------------------|------------------|----------------------------------------------------------------------------------------|
-| vm_compute    | ➕➕➕         | Full          | ❌                         | ❌                       | ❌                | ➕➕➕ Bespoke Cast honored by Qed                                                        |
-| lazy          | ➕➕           | Full / Head   | ❌                         | ➕➕                     | ❌                | ➕➕ Default cast does not store reduction flags                                          |
-| cbv,compute   | ➕             | Full / Head   | ❌                         | ❌                       | ❌                | ➕ Default cast coincides but will use kernel reduction/conversion instead of cbv         |
+| vm_compute    | ✅✅✅         | Full          | ❌                         | ❌                       | ❌                | ✅✅✅ Bespoke Cast honored by Qed                                                        |
+| lazy          | ✅✅           | Full / Head   | ❌                         | ✅✅                     | ❌                | ✅✅ Default cast does not store reduction flags                                          |
+| cbv,compute   | ✅             | Full / Head   | ❌                         | ❌                       | ❌                | ✅ Default cast coincides but will use kernel reduction/conversion instead of cbv         |
 | hnf           | ❔             | Head          |  ❔                        |  ❔                      |  ❔               |  ❔                                                                                       |
-| simpl         | ➖➖           | Full / Head   | ➕/➖                      | ➖ (only head)           | ➕                | ➖ Uses default cast                                                                      |
-| cbn           | ➖➖➖         | Full / Head   | ➕➕                       | ➕➕                     | ➕➕              | ➖ Uses default cast                                                                      |
+| simpl         | ⛔⛔           | Full / Head   | ✅/⛔                      | ⛔ (only head)           | ✅                | ⛔ Uses default cast                                                                      |
+| cbn           | ⛔⛔⛔         | Full / Head   | ✅✅                       | ✅✅                     | ✅✅              | ⛔ Uses default cast                                                                      |
 | **Proposals** |
-| [kred](#kred) | ➕             | Full / Head   | ➕➕                       | ➕➕                     | ➕                | ➖ Uses default cast                                                                      |
-| [MSP](#msp)   | ➕➕ (= lazy)  | Any (quoting) | ❌ (not necessary)         | ❌ (not necessary)       | ❌                | ➕➕➕ Fully integrated into kernel reduction/conversion                                  |
+| [kred](#kred) | ✅             | Full / Head   | ✅✅                       | ✅✅                     | ✅                | ⛔ Uses default cast                                                                      |
+| [MSP](#msp)   | ✅✅ (= lazy)  | Any (quoting) | ❌ (not necessary)         | ❌ (not necessary)       | ❌                | ✅✅✅ Fully integrated into kernel reduction/conversion                                  |
 
 ## Details
 
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -199,7 +199,7 @@ to blocked sub-sub-terms, etc.
 outside of run.
 
 MSP addresses the use cases [Reification with User Terms](#reif) and [Avoiding Term
- Repitition in Proof Terms](#rep) completely and drastically improves on all existing
+ Repetition in Proof Terms](#rep) completely and drastically improves on all existing
  reduction strategies in the precision it offers for reducing specific terms.
 
 ### Advantages
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -20,8 +20,8 @@
     *   Fully normalizes Rocq terms
     *   Translates Rocq terms to OCaml bytecode for efficiency
     *   Generally faster than all strategies listed below
-    *   Translation does have a cost that is offset by automatic caching of the
-        translated bytecode
+    *   Toplevel definitions have their weakly reduced values cached
+    *   Translation does have a cost that is offset by precompiling toplevel definitions
     *   Does not support `Arguments`
     *   Does not support any reduction flags (not even `head`)
     *   Does not support fixpoint refolding
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -108,6 +108,7 @@ once in checking the type of `f` and once in checking the return type.
 ```coq
 let n := 1 + 1 in
 forall (f : Fin.t n), Fin.t n
+```
 
 Unfortunately not all values can be forced and types such as telescopes will
 produce terms of cubic size without any way to share their reduction.
--- a/text/XXX-reduction.md
+++ b/text/XXX-reduction.md
@@ -86,7 +86,7 @@ perform changes to the shallowly embedded iris goal.
 ### Requirements
 
 *   Must only reduce computation introduced by the automation tactics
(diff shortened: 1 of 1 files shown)


text/XXX-reduction.md
```diff
@@ -0,0 +1,227 @@
...
+| [MSP](#msp)   | ✅✅ (= lazy)  | Any (quoting) | ❌ (not necessary)         | ❌ (not necessary)       | ❌                | ✅✅✅ Fully integrated into kernel reduction/conversion                                  |
+
+## Details
+
+*   `vm_compute`
+    *   Fully normalizes Rocq terms
+    *   Translates Rocq terms to OCaml bytecode for efficiency
+    *   Generally faster than all strategies listed below
+    *   Translation does have a cost that is offset by automatic caching of the
+        translated bytecode
```
silene:
```suggestion
    *   Toplevel definitions have their weakly reduced values cached
    *   Translation does have a cost that is offset by precompiling toplevel definitions
```

text/XXX-reduction.md
```diff
@@ -0,0 +1,227 @@
...
+
+The current MSP prototype has already been tested on examples representing these
+use cases and it is able to overcome existing shortfalls both in complexity of
+use as well as in performance.
+
+### Disadvantages
+
+MSP must live in the kernel for it to be useful at Qed time. It might be
+possible to make it opt-in but that hardly changes anything.
+
```
silene, line 215:
```suggestion

It is incompatible with `vm_compute` and `native_compute`.

```

Janno: Thanks for looking at the draft!

I just remembered and re-read your comment in https://github.com/coq/coq/issues/13881#issuecomment-1513336976. I suppose there really is no way to return unreduced terms from `vm_compute`, is there?

In any case, I'll probably try to expand on "incompatible" a little bit. It seems to me that at least some vm casts could be preserved in terms that use MSP with enough care.

silene: Currently no, but that does not mean that it is that farfetched.

First, the VM is already able to manipulate and return axioms, existential variables, and free variables (which are fully normalized values, but from an operational point of view, they are fully unreduced terms). So, extending accumulators to support arbitrary Gallina term is trivial. One has to be a bit careful with open terms so that bound values are properly kept around. But, accumulators are already able to accumulate information (hence their name), so no major change here either.

Second, if the reduction is only delayed rather than just fully blocked, one has to make sure that the bytecode is still around to restart the reduction. But again, the VM already supports similar patterns, e.g., for the `return` clause of `match` constructs. The kernel even has some mechanisms to ensure that Gallina terms and their corresponding bytecode stay synchronized, because this is needed when caching weakly-reduced toplevel definitions (though, this fragile piece of code is probably not exercised as much as it should be).

The hardest part will be to make sure that the Gallina term in the accumulator and the corresponding bytecode agree on what the bound variables are. Currently, there is absolutely no similar mechanism in the bytecode compiler. To give a concrete example, consider
```coq
Definition foo x := let y := x in block (x + y).
```
The blocked Gallina term has two bound variables (`x` and `y`), while the corresponding bytecode has only one bound value `x` (`y` was optimized away by the compiler).  Restarting the reduction by pushing two values on the VM stack would quickly lead to a memory corruption.

Janno: Thank you very much for the detailed explanation. Are there already plans/ideas for handling the binder discrepancies?

Somewhat orthogonal: It just occurred to me while trying to explain this to coworkers that both `vm_compute` and `native_compute`  are currently trivially compatible with MSP. MSP must not (and does not) change the normal forms of terms and both `vm_compute` and `native_compute` perform full normalization. The only difference from MSP is in the result of weak head reduction (and even there it only changes the shape of arguments to the head constructor/constant/etc, not to the head itself). I should expand on this a bit in this document to avoid confusion. 

So until `vm_compute`/`native_compute` supports weak head reduction I don't think they are affected at all. All we need is to treat the primitives as boxing/unboxing operators of some box type.

silene: > Are there already plans/ideas for handling the binder discrepancies?

No. In fact, last time I suggested that the bytecode compiler should handle let-in constructs in a slightly more predictable way, the idea did not live long: https://github.com/coq/coq/issues/13234

aspiwack: > MSP must not (and does not) change the normal forms of terms and both `vm_compute` and `native_compute` perform full normalization

If this is the intention, then support un `vm_compute` is Very Easy™ (I don't remember how native compute works, but I don't see why it would be a problem). You simply reserve one extra tag, and otherwise compile `block t` like `fun x => t` (better: compile it as a 0-argument closure), so that `block t` is a weak-head normal form. When it comes to reconstruction, force the closure and evaluate `t` (which is exactly what you do with functions). This should let you control the order of reduction as you need (I say “should” because your document doesn't really specify the semantics of `block`/`unblock`/`run` (it should!)).

You need to use a different tag for `block t` than for `fun x => t` because it's going to behave differently (_e.g._ the normal form of `fun x => t` is `fun x => t'`, but if I understand correctly, the normal form of `block t` is `t'` (in both case I take `t'` to be the normal form of `t`)).

Supporting weak-head reduction in the VM is essentially the same problem as @silene describes above. Normalisation-by-evaluation really is a surprising sweet spot: you simplify the problem a lot when you're only asking to reconstruct normal-form terms.

Janno: > (I say “should” because your document doesn't really specify the semantics of block/unblock/run (it should!)).

Sorry for that. I actually opened this PR way too early by accident and that's why the document is still in pretty rudimentary shape. The actual idea was to pass around a draft (in the form of a PR against a fork of this repo, not this repo directly) in order to collect additional challenging use cases for reduction before we (possibly) resume work on discussing/specifying/implementing MSP (and `kred`). But I somehow ended up targeting the wrong repo and now I am getting super useful feedback here so I really want keep this PR open.

I'll update the document with some more details on MSP next week. For now, the best summary is [the old prototype](https://github.com/coq/coq/pull/17839) itself. The PR description is probably slightly outdated but the code was very close to the desired semantics. The primitives are in [theories/Force/Force.v](https://github.com/coq/coq/pull/17839/files#diff-81b5a5d0ae05328a993c7de07a67d89f3b491d9b3394b61d42ae5ffc619022e1) and there are tests that should be helpful to illustrate the intended spec in [test-suite/success/force.v](https://github.com/coq/coq/pull/17839/files#diff-36f601b239b060e77c35ab71908060a6f7b988b58a29e590ef1f4cb1e85a9668)

Your description of supporting MSP in the absence of weak-head reduction sounds correct to me except that `block t` simply reduces to `block t'`. `@block T t` has type `Blocked T` which can be thought of a record containing `T`. Both `unblock` and `run` function as projections in this simplified view. I don't remember the exact reason but `Blocked T` cannot be convertible to `T`. It probably involves canonicity.

> This should let you control the order of reduction as you need

I really wonder if this could end up helping simplify the discussion around bytecode compilation/optimization. I see that https://github.com/coq/coq/issues/13234 contains strong arguments for pattern matching being the only viable solution for predictable reduction performance and that also matches my personal experience (although I can almost never use `vm_compute` so I don't know much about the performance pitfalls there). MSP would indeed be one possible way to specify the order of reduction even in `vm_compute`. If that could help simplify the bytecode compilation by making some "optimizations" unnecessary, maybe it could also make implementing weak-head reduction easier in the long run by keeping the bytecode very close to the original terms. (Weak-head reduction in `vm_compute` would be incredibly useful independently of anything discussed here.)

text/XXX-reduction.md
```diff
@@ -0,0 +1,227 @@
...
+|---------------|----------------|---------------|---------------------------|--------------------------|------------------|----------------------------------------------------------------------------------------|
+| vm_compute    | ✅✅✅         | Full          | ❌                         | ❌                       | ❌                | ✅✅✅ Bespoke Cast honored by Qed                                                        |
+| lazy          | ✅✅           | Full / Head   | ❌                         | ✅✅                     | ❌                | ✅✅ Default cast does not store reduction flags                                          |
+| cbv,compute   | ✅             | Full / Head   | ❌                         | ❌                       | ❌                | ✅ Default cast coincides but will use kernel reduction/conversion instead of cbv         |
+| hnf           | ❔             | Head          |  ❔                        |  ❔                      |  ❔               |  ❔                                                                                       |
```
JasonGross:
```suggestion
| hnf           | ❔             | Head          |  ❔                        |  ❌                      | ✅                |  ❔                                                                                       |
```
Last I checked, hnf uses the reduction strategy and refolding engine of simpl. https://github.com/coq/coq/blob/c9f120d29d3f4c8a6a1cf53a734eb2bc7186cac0/doc/tools/docgram/fullGrammar#L800 indicates that it does not support flags.

Janno: Thanks!

JasonGross:
Another reduction use case: 
- reducing PHOAS passes (must be fast, integrate with Qed, and have the ability to force normalization under lambdas before later applying them) --- currently my solution is to convert to de Bruijn and back

Janno:
> currently my solution is to convert to de Bruijn and back

Does that mean you replace terms by de Bruijn indices, reduce, and then substitute back the original terms?

text/XXX-reduction.md
silene: > MSP would indeed be one possible way to specify the order of reduction even in `vm_compute`. If that could help simplify the bytecode compilation by making some "optimizations" unnecessary

MSP does not make the above optimization unnecessary. (The only purpose of this optimization is to allow the garbage collector to free memory earlier in some very rare cases. This optimization is not even part of OCaml's bytecode compiler, for good reasons. It is specific to Coq.) And except for peephole ones, there is no other optimization in the bytecode compiler anyway. So, MSP and bytecode optimizations are two completely separate issues.

JasonGross:
I have a pass that converts from PHOAS representation to de Bruijn representation and a pass that converts from de Bruijn representation to PHOAS, and I compose these passes to force reduction under lambdas.  (Though there's an additional benefit to doing it this way, which is that this also works in extracted code, which an alternative reduction strategy would not.)

MSoegtropIMC:
@Janno: thanks for the initiative and outing together the overview.

I have similar "reification with user terms" problems as you described above. I solve this with an Elpi program which produces positive symbol lists via symbol set operations (take all symbols from modules X Y z, but exclude this or that). This is easy to manage and reliable. In general I think it would make sense to have symbol set operations build into Coq.

An issue not mentioned here is using Coq by non experts and hiding everything which is not at the "user level", which is defined by a well defined set of symbols. Positive list reduction doesn't work here. Imagine I have Z in my "user level" but not positives. If I have a computation in Z which results in a literal Z I would like to do the computation. But this usually requires expanding Pos functions. One more wants a positive list of what is allowed in the final result, not on what is reduced. I don't have a good algorithm in mind for this. One way would be to set a backtracking point when one introduced symbols not in the user level and go back there if the final result is not in the user level. But I guess one could speed this up substantially by storing some meta informations on what king of symbols a reduction of a definition might contain. This also requires some sort of symbol set handling in Coq.

For debugging reduction it would help if one could ask which symbols are contained in the result instead of printing the term (a display option like "Set Printing Symbols" with a few options like naming only the modules or user defined symbol sets involved.

text/XXX-reduction.md
```diff
@@ -0,0 +1,227 @@
...
+# Problematic Reduction Examples
+
+## <a id="reif"></a> Reification with User Terms
+Iris Proof Mode and BlueRock’s automation use reflection and computation to
+perform changes to the shallowly embedded iris goal.
+
+### Requirements
+
+*   Must only reduce computation introduced by the automation tactics
+    *   Can be addressed with whitelist reduction flags, or
```
gmalecha: Alone, this isn't really really enough. You really need an **and** here since functions that need to be reduced often use parts of the standard library such as list append and such.

```suggestion
    *   Can be addressed with whitelist reduction flags, **and**
```

text/XXX-reduction.md
```diff
@@ -0,0 +1,227 @@
...
+
+This approach works wonders for term size but Qed performance depends entirely
+on the ability to force terms across the forall boundary of this function type
+of dynamic arity. For example, in the following type, `1 + 1` is reduced twice:
+once in checking the type of `f` and once in checking the return type.
+
+```coq
+let n :=
+1 + 1 in forall (f : Fin.t n), Fin.t n
+```
```
gmalecha:
```suggestion
```coq
let n := 1 + 1 in
forall (f : Fin.t n), Fin.t n
```
```

File: text/XXX-reduction.md [hunk 6]
```diff
@@ -0,0 +1,227 @@
...
+
+blocked terms may contain unblocked sub-terms which are going to be reduced up
+to blocked sub-sub-terms, etc.
+
+`run` is what triggers MSP to work. `block` and `unblock` have no meaning
+outside of run.
+
+MSP addresses the use cases [Reification with User Terms](#reif) and [Avoiding Term
+ Repitition in Proof Terms](#rep) completely and drastically improves on all existing
```
gmalecha:
```suggestion
 Repetition in Proof Terms](#rep) completely and drastically improves on all existing
```

PR #107, opened 2025-06-04
Asserting outputs of vernacular commands and tactics

radrow:
Render: <https://github.com/radrow/rfcs/blob/vernact-output-assert/text/000-vernac-output-assert.md>

Referenced issue: rocq-prover/rocq#20688

### DIFF
--- a/text/000-vernac-output-assert.md
+++ b/text/000-vernac-output-assert.md
@@ -67,23 +67,6 @@ f = fun x : nat => x + 3
 Print x.
 ```
 
-It should apply to tactics in proof mode as well:
-
-```coq
-Goal False.
-  With Output "
-   (* debug trivial: *)
-    * assumption. (*fail*)
-    * intro. (*fail*)
-  "
-  debug trivial.
-
-  With Output "
-    The command has indeed failed with message:
-    The type has no constructors.
-  "
-  Fail constructor.
-```
 
 In the first iteration, the contents of the string literal must perfectly match
 the command output modulo duplicated whitespaces.
--- a/text/000-vernac-output-assert.md
+++ b/text/000-vernac-output-assert.md
@@ -1,0 +1,140 @@
+- Title: Assert outputs of vernacular commands and tactics within Rocq code
+
+- Drivers: Radosław Jan Rowicki @radrow
+
+----
+
+# Summary
+
+I would like to see and test outputs of vernacular commands without necessarily
+going into interactive. More concretely, I propose to add a feature that lets me
+paste the desired output of a particular command into Rocq code and make Rocq
+verify whether it matches. The concept is similar to the existing `Fail`
+command.
+
+This is by the way a
+[feature](https://florisvandoorn.com/carleson/docs/Lean/Elab/GuardMsgs.html) in
+Lean (the docs are quite poor though).
+
+# Motivation
+
+### Testing
+
+First, compiler and plugin testing. Right now
+[rocq-prover/rocq](https://github.com/rocq-prover/rocq) has some awkward tests
+that store reference outputs in separate files which are then compared to
+produced outputs from `.v` samples. This feature would allow writing such tests
+more clearly and without additional scripting.
+
+Second, this aids testing Rocq code. Currently there exist workarounds for
+particular cases, but there is no universal way to test things like `Check` on
+modules, `Compute`, or `Ltac` scripts. A good example of this is the `Fail`
+command which currently allows only checking that a command has failed, but not
+the reason of the failure. Sure, the outputs may change occasionally and require
+manual intervention, but that doesn't happen that often due to Rocq updates, and
+if it's due to changes in code, then it's not a shock that you have to update
+the test as well. Also fixing it is not really a big deal. Rocq and IDEs could
+even support autoupdating reference outputs.
+
+### Presentation
+
+Having the ability to recall a definition makes Rocq scripts much more readable,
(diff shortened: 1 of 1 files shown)


text/000-vernac-output-assert.md
```diff
@@ -0,0 +1,139 @@
...
+    * intro. (*fail*)
+  "
+  debug trivial.
+
+  With Output "
+    The command has indeed failed with message:
+    The type has no constructors.
+  "
+  Fail constructor.
+```
```
SkySkimmer:
```suggestion
```
tactics are commands so this part is trivial

text/000-vernac-output-assert.md
```diff
@@ -0,0 +1,139 @@
...
+
+In the first iteration, the contents of the string literal must perfectly match
+the command output modulo duplicated whitespaces.
+
+### Possible improvements
+
+Just ideas. To be discussed in more detail after this RFC is concluded.
+
+1. Allow regexes.
+2. Flag to turn this off — both in code and as a CLI parameter.
```
SkySkimmer: if we produce a warning instead of error on mismatch ths is trivial
(note that warnings can be errors by default, unsupported-attributes is an example)

text/000-vernac-output-assert.md
```diff
@@ -0,0 +1,139 @@
...
+3. We can deduct how to parse output of some commands and allow more
+   flexibility. For example, we could accept some minor term conversions in
+   cases of `Check` and `Print`.
+
+# Drawbacks
+
+- Another restricted keyword (`With`).
```
SkySkimmer: I don't think it needs to be restricted (eg Fail isn't)

Alizter:
This will act similarly to ppx expect for OCaml I suppose? https://github.com/janestreet/ppx_expect

radrow:
> This will act similarly to ppx expect for OCaml I suppose? https://github.com/janestreet/ppx_expect

Pretty much. That looks like a good source of further ideas.

gares:
I like this proposal. I prefer Assert Output rather than With Output, or #[expected(output="...")].

ybertot:
I think the proposal can also be very useful to write tutorials.  I have two improvements to suggest:

 - the expected output should appear in the script after the command that is concerned, so that the flow of reading looks more chronological
 - It would be nice to be able to include a filter, a unix command that treats the both output string and the expected text and removes irrelevant information (I used such a filter in the Coq'Art book, where the aim was to be independent from spacing and indentation).

radrow:
> the expected output should appear in the script after the command 

I'd also prefer the output to go after the command, but I can't come up with nice syntax for it (e.g. so that the desired output doesn't look like a parameter to the command and especially isn't confused with optional parameters). Maybe the command could be wrapped with some tokens, but that might look weird as afaik no existing commands use such syntax. You can throw ideas if you have any

SkySkimmer:
Maybe we could separate into 2 commands, looking like
~~~coq
Capture Output Check foo.
Assert Output "foo : nat".
~~~

The original issue had
~~~coq
Check foo.
Output "foo : nat".
~~~
ie implicit capture with the assert applying to the previous command, but this wouldn't work well with eg proof general inserting arbitrary commands for itself between the user's commands.

With explicit capture we can insert non-captured commands without issue:
~~~coq
Capture Output Check foo.
Check bar.
Assert Output "foo : nat".
~~~

we can also combine outputs:
~~~coq
Capture Output Check foo.
Capture Output Check bar.
Assert Output "
foo : nat
bar : bool
".
~~~

To prevent forgetting outputs, if there is a Capture Output but no Assert Output before the end of the file compilation fails (like with leaving open proofs). We could also fail if there is no Assert Output before a End (not sure if desired).


>It would be nice to be able to include a filter, a unix command that treats the both output string and the expected text and removes irrelevant information

I don't think we want to be able to call unix commands from Rocq code (security issues, not portable...).
We should probably normalize whitespace by default (maybe with an option to disable normalization) (eg Printing Width depends on user settings so would make things non reproducible).

We could also have a way to call Rocq code on the output string (ie expose it as a primitive string) somehow.
eg `Process Output msg Using code.` would bind `msg`  to the current output, evaluate `code` and set the result as the new output.
`code` could be a regular rocq term eg `PrimString.sub msg 0 4` to get the first 4 chars, or it could use tactics in terms eg `ltac2:(do_stuff msg)`.

ybertot:
I like the `Capture/Assert` pair.  I understand the reluctance to depend on platform dependent filters.

radrow:
That's a great idea, I really like the ability to combine two outputs. To that I would also add 

- `Drop Output.` — if you change your mind (useful in `coqtop`). 
- `Show Output.` — to print the contents of the output (`Print` would make clashes)

Also a fun trick: if you want to assert output without flushing it, you can do `Fail Fail Assert Output`.

This could be later extended to more advanced output caching with assignable buffers that can be managed more precisely (dumping to files, combining, whatever).

TheoWinterhalter:
Note that `Fail Fail` can be replaced by `Succeed` already.

SkySkimmer:
https://github.com/rocq-prover/rocq/pull/22140

cpitclaudel:
A side note since I missed most of this discussion: Alectryon already has a feature to assert that a particular command produced a certain output, along with a DSL to select subparts of the goal to assert things about: https://github.com/cpitclaudel/alectryon/blob/main/recipes/references.rst#asserting-properties-of-the-output

PR #109, opened 2025-07-02
Multiple Default Typeclass Databases

Janno:
Rendered: https://github.com/janno/ceps/blob/hint-db/text/XXX-hint-db.md

### DIFF
--- a/text/XXX-hint-db.md
+++ b/text/XXX-hint-db.md
@@ -49,7 +49,7 @@ The key observation that motivates this CEP is that the problem is almost
 entirely trivial to solve in many cases. In particular, many developments
 already work on entirely or almost entirely disjoint typeclasses. The only
 reason they all end up in the same database is because the system is set up in a
-way that makes that choice the past of least resistance (and sometimes the only
+way that makes that choice the path of least resistance (and sometimes the only
 path, e.g. if clients are expected to benefit from the typeclasses in `apply _`
 and the other tactics listed above).
 
--- a/text/XXX-hint-db.md
+++ b/text/XXX-hint-db.md
@@ -1,0 +1,160 @@
+- Title: Multiple Default Typeclass Databases
+
+- Drivers: @Janno, @mattam82
+
+----
+
+# Summary
+
+The hint databases `typeclass_instances` is special in ways that cannot be
+emulated by using other databases. The user experience of building up class
+hierarchies and proof automation outside the default database is practically
+prohibitive. This forces most Rocq developments to share a single database, and,
+thus, makes it all but impossible to fine-tune settings such as opacity to the
+needs of any individual project unless the project is never combined with other
+projects that use typeclasses.
+
+We propose to allow projects to register new "default TC databases", i.e.
+databases that are used by default in tactics such as `apply` and `rewrite`. By
+disentangling the identity of the database(s) used from their status as a
+default TC database, we allow projects to regain full control over all hint database
+settings.
+
+# Motivation
+At the time of writing, `typeclass_instances` is the only database that is used
+seriously for developing typeclass hierarchies and proof automation in almost
+all Rocq developments. This is owed to its special status in tactics and certain
+vernacular commands:
+
+1. At least `apply _`, `typeclasses eauto` (without arguments), and
+   `(setoid_)rewrite` hardcode `typeclass_instances` as the (only) source of instances.
+2. `typeclass_instance` is the default database for opacity and hint mode information for instances found in the local context.
+3. `Instance` is restricted to `typeclass_instances` (`Hint Resolve` exists as a partial workaround).
+4. `Typeclasses Opaque/Transparent` is restricted to `typeclass_instances` (`Hint Opaque/Transparent` exists as a workaround).
+
+All these restrictions, but mostly (1), severely harm modularity: by forcing all
+Rocq developments to go through a single database, Rocq makes it impossible to
+enforce project-specific database settings that make sense for one development
+but not for others.
+
+This is most visible in opacity settings: for example, two developments that
+make use of `List.map` in their typeclass searches might have conflicting needs
+about its matching and reduction behavior during the search. One client might
+need to have the function reduce and thus set it to be transparent, while
+another client might rely on it being opaque for performance reasons. This is
+currently impossible if these two developments want to be able to co-exist,
+possibly in a third development that includes them both.
+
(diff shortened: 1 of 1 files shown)


text/XXX-hint-db.md
```diff
@@ -0,0 +1,160 @@
...
+need to have the function reduce and thus set it to be transparent, while
+another client might rely on it being opaque for performance reasons. This is
+currently impossible if these two developments want to be able to co-exist,
+possibly in a third development that includes them both.
+
+The key observation that motivates this CEP is that the problem is almost
+entirely trivial to solve in many cases. In particular, many developments
+already work on entirely or almost entirely disjoint typeclasses. The only
+reason they all end up in the same database is because the system is set up in a
+way that makes that choice the past of least resistance (and sometimes the only
```
rlepigre-skylabs-ai:
```suggestion
way that makes that choice the path of least resistance (and sometimes the only
```

rlepigre-skylabs-ai (commented):
This proposal seems extremely useful to me!

Having a vernacular command that controls what database is used by default by `Instance` and `Typeclasses Transparent`/`Opaque` is probably a must have to more easily port projects to using a specific database.

MackieLoeffel:
Thanks for this proposal! I think this would be very useful for the proof automation I am working on. I've often encountered the problem that I would like to use multiple different TC databases, but it is just very annoying to do this currently.

How would this CEPS affect the `Instance` command? Would the instances be added all default databases? Or would it be prohibited to use `Instance` without specifying a database? (The text currently reads like it might be the second, but it is not entirely clear.)

Janno:
> How would this CEPS affect the `Instance` command? Would the instances be added all default databases? Or would it be prohibited to use `Instance` without specifying a database? (The text currently reads like it might be the second, but it is not entirely clear.)

The proposal contains a small remark at the end about a way to select the default TC database for commands such as `Instance` (https://github.com/janno/ceps/blob/hint-db/text/XXX-hint-db.md#possibly-select-default-typeclass-database). I do not think any command that mutates TC database state should mutate the entire set of registered databases. I can't think of a situation where that would be the desired outcome. Given that @rlepigre-skylabs-ai also asked about this in his comment above I think I will amend the proposal and make the command/setting to select which database these commands act on a required part of the proposal instead of just a possibility.

robbertkrebbers:
I am not a fan, this seems like a feature to work around a "bug".

In the long run, it seems like a better idea to make sure to have more principled options for type classes, rather than a zoo of options where everyone can make their own database to configure these options differently.

What is the competition doing? AFAIK Haskell just has a single database for type classes. But what about Lean?

This solution also seems to make the diamond problem even worse. Now you have a development `X` that depends on two libraries `L1` and `L2`, that both have their own database. To which database is `X` supposed to add its instance? The database of `L1` or `L2`, or both?

Janno:
> In the long run, it seems like a better idea to make sure to have more principle options for type classes, rather than a zoo of options where everyone can make their own database to configure these options differently.

I am not sure that this RFC addresses the problems you are thinking of. Are you thinking of a more fundamental redesign of the typeclass system? With a single database used by everyone it seems to me that there is no way to reconcile the needs of all clients. One development might treat `List.map` as an opaque object about which it registers instances (e.g. `Proper`) while another one relies on the same `List.map` function to be transparent because it needs to be unfolded during TC unification.

> What is the competition doing? AFAIK Haskell just has a single database for type classes. But what about Lean?

I am not familiar enough with Haskell to comment on that in particular. I assume they do not have unification with delta in their typeclass searches but I could be wrong. Lean is much closer to Rocq and they do have some defaults that are better than ours. For example, class projections are treated as opaque by default. Apart from that, Lean's support for tabling probably helps to avoid penalties from lax opacity settings but they might still end up suffering down the road. It is hard to tell what will happen given that Lean simply has not been around that long. I suspect they also have better metaprogramming and tactic support than Rocq. That could lead people away from relying on typeclass search for automation purposes (which is where most of our problems show up).

> This solution also seems to make the diamond problem even worse. Now you have a development X that depends on two libraries L1 and L2, that both have their own database. To which database is X supposed to add its instance? The database of L1 or L2, or both?

This is a good point that is not yet addressed in the proposal. Given that deduplication of hints across databases is not generally sound I don't think adding the hint to both databases is generally a good choice. So this leaves either the L1's or L2's database. If the new instances are of an existing class then it is likely only present in one database and that's the one to pick in almost all cases. If the class is new but its instances rely on classes from both databases then the user is free to pick whichever database has the most fitting transparency settings for their own instances. Or make a new database with exactly the right settings.

Making a new database will possibly involve copying a bunch of settings from either L1, L2, or both. There is a whole can of worms to open here about an algebra of databases that could allow people to define a database in terms of one or more existing databases by copying and possibly unifying existing ones (with careful choices about conflicting settings). They could then remove the existing DBs from the set of default TC databases and register only the new one. I think this would be my ideal solution to the diamond problem. But it is probably outside the scope of this RFC.

robbertkrebbers:
Do I understand that the mode of use you generally have in mind is the following:

When you develop automation based on type classes, you typically define a bunch of new class. Now for those classes you use a new database with your custom settings (e.g., for unfolding).

Isn't it then more the case that for each *class* you want to be able to have a custom database with opacity settings (and other settings)?

Janno:
> When you develop automation based on type classes, you typically define a bunch of new class. Now for those classes you use a new database with your custom settings (e.g., for unfolding).

Yes, that is what I had in mind. 

> Isn't it then more the case that for each class you want to be able to have a custom database with opacity settings (and other settings)?

I suppose the RFC does not preclude that way of using databases. So this could indeed be one way to go. I personally think one db per class seems too much but I could certainly see different subsystems within a development using different databases. It's really not so much about any particular unit such as a class or instance but rather a coherent package of classes and instances that are tuned to work together within one database. But there is no reason it couldn't be just one class.

robbertkrebbers:
I didn't mean that you make a new database for each class, but rather that you associate to each class a database. So multiple classes can be associated to one database, but each class only has a single database.

Another advantage is that you do not need to specify which database to use each time you declare an instance. You only specify that for a class.

Janno:
Oh, I see. I think this might be a great addition to the RFC. Instead of setting a default TC database for all `Instance` commands and the like we could (also) associate the database with the class. I am not sure if we want to forbid hints for one class to appear in multiple database if users truly want that but making the default configurable per class sounds great.

